PR #10, opened 2022-11-08
Sales - GL transformations

amar-fw:


apps/backend/src/resources/fixedWidthRecord/fixedWidthRecord.ts
```diff
@@ -69,7 +72,7 @@ export class FixedWidthRecordService {
    * @param clz the class/constructor
```
chelsea-EYDS, line 72: I realize this is not a part of the diff, but I am wondering if we should include these jsdoc annotations throughout the codebase? This could be useful later for generating documentation. Although, I am not sure if we need/want both jsdoc and open api or if exclusively using one of these would be best?

apps/backend/src/resources/spec/fixtures/sale.json
```diff
@@ -0,0 +1,41 @@
+{
+    "id": "test",
+    "sale_date": "2022-10-12",
+    "journal_name": "SM J000001",
+    "ministry_alpha_identifier": "SM",
+    "total_amount": 50,
+    "distributions": [
+        {
+            "line_number": "00001",
```
chelsea-EYDS, line 9: It looks like the tab width is extra wide here, but possibly this is intentional?

apps/backend/src/resources/Resource.ts
```diff
@@ -0,0 +1,32 @@
...
+  resourceType?: string;
+  transformSalesEvent?(record: SalesDTO, options: transformSalesEventOptions) : T
+}
+
+export type transformSalesEventOptions = {
+    index: number
+};
+
+export class Resource<T extends ResourceBase<T>> implements ResourceBase<T> {
+    public resource: T;
```
chelsea-EYDS, line 14: It seems like there might be some inconsistent formatting on tab-width?

.prettierrc
```diff
@@ -3,7 +3,9 @@
   "trailingComma": "all",
   "arrowParens": "avoid",
   "tabWidth": 2,
+  "useTabs": false,
   "semi": true,
```
chelsea-EYDS, line 7: I wonder if a pre-commit hook might be useful to apply linting and formatting rules?

apps/backend/src/resources/spec/GLRecord.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { GLRecord } from '../';
+import { SalesDTO } from '../../sales/dto/sales.dto';
+
+const salesRecord = require('./fixtures/sale.json');
+describe('GLRecord', () => {
+  it('transform from sales event ', () => {
+    const glRecord = GLRecord.transformSalesEvent([salesRecord as SalesDTO]);
+    console.log(glRecord.batchHeader.convertFromJson().toString());
```
chelsea-EYDS, line 8: Is this log intended for development only? Should we use the custom logger here instead of console?

apps/backend/src/lambdas/generateGL.ts
```diff
@@ -41,60 +30,51 @@ export const handler = async (event?: any, context?: Context) => {
...
-  return convertToFixedWidth(glRecord);
-};
+//   const glRecord = new GLRecord({});
+//   glRecord.jv = [];
+//   const theMap = new Map<
+//     string,
+//     {
+//       header: JVHeader;
+//       details: JVDetails[];
+//     }
```
chelsea-EYDS, line 46: I am assuming this is a TODO? :D

chelsea-EYDS (commented):
Looks good to me! I left a few minor comments but all were related to the formatting. I am going to fetch the branch and test things out to get some better context on the logic.

apps/backend/src/resources/spec/Resource.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { Resource } from '../Resource';
+
+describe('Resource', () => {
+  it('resourceType should  be defined when initializing a Resouce', () => {
+    class TestResource extends Resource<{}> {}
+    try {
+      const object = new TestResource({});
+    } catch (e) {
+        expect(e instanceof Error).toBeTruthy();
+        expect(e.message).toBe('no resource type specified on TestResource');
```
chelsea-EYDS, line 10: Are we using a custom exceptions messages? We might be able to make the "expect" something more specific if we have standard error messages in the case of an empty Test Resource?

chelsea-EYDS (approved):

PR #11, opened 2022-11-09
Precommithook

chelsea-EYDS:
**Changes:**

- Created a Dockerfile to run for local dev in order to keep environment consistent
   - The extended tsconfig from the root dir was not accessible in the container, so I have manually included the original tsconfig options in the app/backend tsconfig file
   - I had a couple of errors/warnings pop up. After installing axios and express in backends/app and making a slight modification in the logger service (see diff) these were resolved. 
   - In app module the endpoint and region for the awsSDK are being set by env vars specified in the docker-compose file. 
  
- Installed husky and added a precommit hook to run eslint + prettier
  - This will run on any commit, but will only check the file types as specified in lint staged. Prettier will be applied according to the .prettierrc file. 
  - To override append "-n" to your commit message
Screen Shot 2022-11-09 at 1 55 07 PM

- Installed eslint nibble into the dev dependencies
  - This is just a nice to have package for easily addressing lint errors/warnings
Screen Shot 2022-11-09 at 1 54 05 PM


apps/backend/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM node:16.18-alpine As development
+
+WORKDIR /usr/src/app
+
+COPY  package*.json ./
```
vyasworks, line 5: do we need this or yarn lock?

vyasworks:
The `.yarn` files should not be deleted

PR #12, opened 2022-11-19
TDI-17/TDI-34 File parse 

chelsea-EYDS:
**What has changed:**

Added class definitions for TDI17/TDI34 files.
Added sample files to a root directory (could not access sharepoint files from the terminal)
Added Makefile commands to parse local. Parsing requires a file path and file type, either passed in the event, or, set as env vars. Env vars are set for local dev in makefile cmd.


apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+import { TDI17Record } from '../tdi-parser/tdi17/TDI17Record';
+import { TDI17Header } from '../tdi-parser/tdi17/TDI17Header';
+import { TDI17Details } from '../tdi-parser/tdi17/TDI17Details';
+import { TDI17Trailer } from '../tdi-parser/tdi17/TDI17Trailer';
+import { TDI34Record } from '../tdi-parser/tdi34/TDI34Record';
+import { TDI34Header } from '../tdi-parser/tdi34/TDI34Header';
+import { TDI34Details } from '../tdi-parser/tdi34/TDI34Details';
+import { TDI34Trailer } from '../tdi-parser/tdi34/TDI34Trailer';
+import { TDI34Enum } from './const/tdi34.const';
+import { TDI17Enum } from './const/tdi17.const';
```
vyasworks, line 15: could be imported as a single module for 17 and 34 each.

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+  context?: Context,
+) => {
+  const app = await NestFactory.createApplicationContext(AppModule);
+  const appLogger = app.get(AppLogger);
+  const s3manager = app.get(S3ManagerService);
+
+  appLogger.log({ event });
+  appLogger.log({ context });
+
+  const filetype = event?.type ?? process.env.TDI_TYPE;
```
vyasworks, line 28: Suggestion - This line should throw an error if the event type does not exist. Defaulting should not be an option

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+  const TDIEnum = filetype === 'TDI17' ? TDI17Enum : TDI34Enum;
+
+  try {
+    appLogger.log('...start TDI Parsing');
```
vyasworks, line 32: NIT: parsing of which TDI, we could include the Number

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+    const contents = await s3manager.getContents(
+      TDIEnum.BUCKET,
+      event?.filePath ?? TDIEnum.LOCAL_PATH,
```
vyasworks, line 35: we should check for this path along with the event type in the initial processing script. 

sort of validation block. 

we could use - https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+    );
+
+    const record = contents?.Body?.toString() || '';
```
vyasworks, line 38: why do we default to an empty string here?

apps/backend/src/lambdas/const/tdi34.const.ts
```diff
@@ -0,0 +1,10 @@
+export enum TDI34Enum {
+   BUCKET= 'bc-pcc-data-files-local',
+   LOCAL_PATH= 'tdi34/TDI34.TXT',
+   OUTPUT_PATH='outputs/tdi34/',
+   JSON_FILENAME='_TDI34.json',
+   JSON_TYPE='/application/json',
```
vyasworks, line 6: is`JSON_TYPE` necessary to be in the enum?

apps/backend/src/lambdas/const/tdi34.const.ts
```diff
@@ -0,0 +1,10 @@
+export enum TDI34Enum {
+   BUCKET= 'bc-pcc-data-files-local',
+   LOCAL_PATH= 'tdi34/TDI34.TXT',
```
vyasworks, line 3: I don't think we should have local file path or any logic for that. 

we can enable testing scripts to work with local stack.

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+    const output = parseTDI(record, TDIEnum);
+
+    appLogger.log(output);
+    const outputFile = Buffer.from(JSON.stringify(output));
+
+    appLogger.log(outputFile);
+    await s3manager.putObject(
```
vyasworks, line 45: I would prefer this to have it's own try catch.

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+};
+
+const parseTDI = (fileContents: string, TDIEnum: unknown) => {
+  const TDI_17 = TDIEnum === TDI17Enum;
+
+  const lines = fileContents.split('\n').filter((l: string) => l);
+
+  const header = TDI_17 ? new TDI17Header({}) : new TDI34Header({});
+  const trailer = TDI_17 ? new TDI17Trailer({}) : new TDI34Trailer({});
```
vyasworks, line 62: I don't think we want to store the header and trailer for TDI files. 

The transactions line items is all we want.

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  record.details = TDI_17 ? tdi17RecordDetails : tdi34RecordDetails;
+  record.trailer = trailer;
+  record.header = header;
+
+  return {
+    header: record.header.resource,
+    details: record.details.map((itm: { resource: unknown }) => itm.resource),
+    trailer: record.trailer.resource,
+  };
```
vyasworks, line 96: we need only the array of details.

apps/backend/src/tdi-parser/tdi17/TDI17Details.ts
```diff
@@ -0,0 +1,200 @@
+import {
```
vyasworks, line 1: create folder level index files and export from them.

apps/backend/src/tdi-parser/tdi17/TDI17Header.ts
```diff
@@ -0,0 +1,121 @@
+import {
```
vyasworks, line 1: We could skip the header and trailer as we don't have any use for it  YET. So we can leave it here and use if we need it later.

vyasworks (changes requested):

apps/backend/src/lambdas/const/tdi34.const.ts
chelsea-EYDS: Agreed, this is actually a reference to the file in localstack. In order to develop locally I am calling `make upload-local-tdi(17||34)` which does rely on the static file in sample-files firectory at the root. `LOCAL_PATH` should be `LOCALSTACK_PATH` ?

apps/backend/src/tdi-parser/tdi17/TDI17Header.ts
chelsea-EYDS: Okay sounds good. I thought we might want the file header as there is the file creation date. So for now I will add an index, and export only the details for 17/34?

apps/backend/src/lambdas/const/tdi34.const.ts
vyasworks: When we call the make command, we could also pass the local stack's path. 

This way anyone who wants to test and play with different data sets is not forced to use one file path and keep overwriting it. `tdi34/TDI34.TXT`

apps/backend/src/tdi-parser/tdi17/TDI17Header.ts
vyasworks: Yes. 

The date seems to also be in each of the transaction items, so it might just be redundant. 
Is the header date different or providing us any additional context? 🤔

Makefile
```diff
@@ -141,3 +141,23 @@ deploy-api:
 
 deploy-gl:
 	aws lambda update-function-code --function-name glGenerator --zip-file fileb://./.build/pkg/backend.zip --region $(AWS_REGION) > /dev/null
+
+# ===================================
+# Add Local Tdi Files
+# ===================================
+
+put-local-tdi17:
```
vyasworks, line 149: the make command can also accept paths

vyasworks: https://stackoverflow.com/a/2826069

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,114 @@
...
+    const contents = await s3manager.getContents(TDI_FILE.BUCKET, filepath);
+
+    const record = contents?.Body?.toString();
+    const output = record && parseTDI(record, TDI_FILE);
+
+    appLogger.log(output);
+
+    await uploadParsedTDI(TDI_FILE, s3manager, output, appLogger);
+  } catch (e) {
+    appLogger.error(e);
```
vyasworks, line 60: we should exit the handler here gracefully.

vyasworks: or pass the error back to the root fn and exit there

vyasworks: It is also okay to not handle it because we will see the error in the console, but more logs and structure will help debugging in the future

apps/backend/src/tdi-files/const.ts
```diff
@@ -0,0 +1,12 @@
+export const generateEnum =(fileType: string)=> {
+  return {
+   TYPE: fileType,
+   BUCKET:  'bc-pcc-data-files-local',
+   LOCALSTACK_PATH:  `${fileType.toLowerCase()}/${fileType.toUpperCase()}.TXT`,
```
vyasworks, line 5: to be removed all together for providing flexible testing

apps/backend/src/tdi-files/const.ts
```diff
@@ -0,0 +1,12 @@
+export const generateEnum =(fileType: string)=> {
+  return {
+   TYPE: fileType,
+   BUCKET:  'bc-pcc-data-files-local',
```
vyasworks, line 4: Bucket name needs to change by ENV

Makefile
vyasworks: example: https://github.com/bcgov/hcap/blob/master/Makefile#L56

Usage: https://github.com/bcgov/hcap#to-seed-the-database-with-data-in-an-excel-sheet

apps/backend/src/lambdas/parseTDI.ts
```diff
@@ -0,0 +1,72 @@
...
+export const handler = async (event?: any, context?: Context) => {
+  const app = await NestFactory.createApplicationContext(AppModule);
+  const appLogger = app.get(AppLogger);
+  const s3manager = app.get(S3ManagerService);
+
+  appLogger.log({ event });
+  appLogger.log({ context });
+
+  if (!event.type || !event.filepath) {
+    throw new BadRequestException();
```
chelsea-EYDS, line 18: throw Error

apps/backend/package.json
```diff
@@ -10,15 +10,17 @@
...
     "start:prod": "NODE_ENV=prod node dist/main",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
     "test": "jest --coverage --verbose --forceExit --runInBand",
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json --runInBand"
+    "test:e2e": "jest --config ./test/jest-e2e.json --runInBand",
+    "parseTDI17": "NODE_ENV=local RUNTIME_ENV=local LAMBDA_ARG=TDI17 ts-node ./scripts/handler.ts",
+    "parseTDI34": "NODE_ENV=local RUNTIME_ENV=local LAMBDA_ARG=TDI34 ts-node ./scripts/handler.ts"
```
vyasworks, line 23: ts-node ./scripts/handler.ts console.log(handler.({type: "TDI34", path: "pcc-bc-xxxx-local"}))

something like this

apps/backend/scripts/handler.ts
```diff
@@ -0,0 +1,7 @@
+import {handler} from '../src/lambdas/parseTDI'
```
vyasworks, line 1: Proabbaly dont need this file

Makefile
```diff
@@ -141,3 +141,23 @@ deploy-api:
...
+
+put-local-tdi34:
+	awslocal s3api put-object --bucket bc-pcc-data-files-local --key tdi34/TDI34.TXT --body ./sample-files/TDI34.TXT 
+
+# ===================================
+# Parse Local
+# ===================================
+
+parse-local-tdi17:
+	cd './apps/backend' && yarn run parseTDI17
```
vyasworks, line 160: node -e 'require(\"./lambdas/parseTDI\").handler({type: "TDI17", path: "/xasxasx/xasxas/xasxas"})'"

vyasworks (approved):

PR #14, opened 2022-11-28
Testrepo

chelsea-EYDS:
**What has changed:**

- Removed root level package.json/node_modules/yarn stuff
- Updated backend tsconfig and jest.config
- created docker build/container for local and test
- updated docker-compose 
- added makefile commands to build/run local and test


.github/workflows/test.yaml
```diff
@@ -0,0 +1,28 @@
+on:
+  push:
+    tags:
+      - dev
+env:
+  TF_VERSION: 1.2.4
```
vyasworks, line 6: remove TF Version

.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
```diff
@@ -1,28 +0,0 @@
-/* eslint-disable */
```
vyasworks, line 1: are we downgrading from yarn3

apps/backend/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+EXPOSE 3000
+
+CMD ["yarn", "run", "start:dev"]
+
+###################
+# BUILD FOR PRODUCTION
+###################
+
+FROM node:16-alpine3.16 AS build
```
vyasworks, line 32: We will run the apps as lambdas, 

but curious why the production build is in the same file as local build

vyasworks (commented):
LG, just few qns. 

Pls clarify what you meant by moving backend one folder up :)

.github/workflows/test.yaml
```diff
@@ -0,0 +1,28 @@
+on:
+  push:
+    tags:
+      - dev
```
vyasworks, line 4: should also run on PRs

apps/backend/Dockerfile
chelsea-EYDS: (I am not sure if I have done the incremental build properly though :)

.github/workflows/test.yaml
chelsea-EYDS: Oops! Thanks!

.github/workflows/test.yaml
chelsea-EYDS: > LG, just few qns.
> 
> Pls clarify what you meant by moving backend one folder up :)

Oops! I un-did this to avoid an even bigger diff. Also, I wasn't sure if the zipped prod build referenced the current file path so I thought it would be better to leave it as is.

vyasworks (approved):

PR #19, opened 2022-12-16
Location Map + SFTP Scripts

chelsea-EYDS:
**What has changed:**

- added scripts to grab bcm and pcc files from sftp servers, parse, and drop in s3 
- slight reorganization of common dir
- fixed imports after implementing last step
- undid sales api update, created a new ticket, moved all previous PR comments to the new ticket
- added location mapping for parsed files


apps/backend/src/common/const/location-dictionary.ts
```diff
@@ -0,0 +1,156 @@
+export const locationToMerchantMap = [
```
vyasworks, line 1: This is great, can we mark this under 'SBC' or add another key ?

apps/backend/src/common/const/location-dictionary.ts
```diff
@@ -0,0 +1,156 @@
...
+  { office: 'Vanderhoof', merchant_terminal: 22044750, location_id: 92 },
+  { office: 'Vernon', merchant_terminal: 22044859, location_id: 93 },
+  { office: 'Victoria', merchant_terminal: 20777441, location_id: 94 },
+  {
+    office: 'Williams Lake',
+    merchant_terminal: 22044933,
+    location_id: 97
+  },
+  {
+    office: 'SBC\u00a0Mobile\u00a0GVA',
```
vyasworks, line 122: Is this unicode character `\u00a0` intentional?

apps/backend/src/flat-files/tdi17/TDI17Details.ts
```diff
@@ -17,10 +17,10 @@ export interface ITDI17Details extends IFixedWidthRecord<ITDI17Details> {
   deposit_time: string;
   seq_no: string;
   location_desc: string;
-  deposit_amt_curr: number;
+  deposit_amt_curr: string;
```
vyasworks, line 20: why is this changing to string?

apps/backend/src/temp-scripts/sftp.garms.sh
```diff
@@ -0,0 +1,16 @@
+# ! /bin/bash
```
vyasworks, line 1: add readme on deets, how to use, export vars, where to find them in our sftp. whom to reach out for pvt keys.

vyasworks (approved):

apps/backend/src/flat-files/tdi17/TDI17Details.ts
chelsea-EYDS: For ease of testing. This way we can ensure the correct number of characters in the string. I had planned to parse it while working with the data later, and to just keep this value exactly as is until we are ready to work with the data.

vyasworks (approved):

PR #21, opened 2022-12-20
Location map 3

chelsea-EYDS:
I have removed the location mapping from the TDI/DDF. Mapping will apply to sales and the corresponding TDI/DDF can be queried using the mapped sales values, rather than a two way mapping on each of the files.


chelsea-EYDS:
No longer relevant - changes on new PR

PR #22, opened 2023-01-04
db and script update

chelsea-EYDS:


apps/backend/src/flat-files/tdi17/TDI17Details.ts
```diff
@@ -13,7 +13,7 @@ export interface ITDI17Details extends IFixedWidthRecord<ITDI17Details> {
   program_cd: string;
   deposit_date: string;
   transaction_type: number;
-  location: string;
+  location_id: number;
```
vyasworks, line 16: Any reason why we switched to number, I think string seems more appropriate here?

apps/backend/src/reconciliation/entities/cash-deposits.ts
```diff
@@ -0,0 +1,52 @@
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+
+@Entity('cash_deposit(TDI17)')
+export class CashDepositEntity {
+  @PrimaryGeneratedColumn({ type: 'int' })
```
vyasworks, line 5: we need to chat more about this.

apps/backend/src/reconciliation/entities/cash-deposits.ts
```diff
@@ -0,0 +1,52 @@
...
+  id: number;
+
+  @Column()
+  program_cd: string;
```
vyasworks, line 9: program_code

apps/backend/src/reconciliation/entities/location.entity.ts
```diff
@@ -0,0 +1,16 @@
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+
```
vyasworks, line 2: convert to local enum as discussed

apps/backend/src/reconciliation/entities/payment-method.entity.ts
```diff
@@ -0,0 +1,10 @@
+import { Entity, PrimaryColumn, Column } from 'typeorm';
```
vyasworks, line 1: convert to local enum as discussed

apps/backend/src/reconciliation/entities/pos-deposit.entity.ts
```diff
@@ -0,0 +1,34 @@
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+
+@Entity('pos_deposit(TDI34)')
+export class POSDepositEntity {
+  @PrimaryGeneratedColumn({ type: 'int' })
+  id: number;
```
vyasworks, line 6: we should chat about ids, thinking how we can make it idempotent?

apps/backend/src/reconciliation/entities/static.entity.ts
```diff
@@ -0,0 +1,31 @@
+import { Entity, Column, PrimaryColumn } from 'typeorm';
+
+@Entity('trans_cd')
+export class TransactionCDEntity {
```
vyasworks, line 4: all of these become enums I gues?

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    @InjectRepository(TransactionEntity)
+    private salesRepo: Repository<TransactionEntity>,
+    @InjectRepository(POSDepositEntity)
+    private TDI34: Repository<POSDepositEntity>,
+    @InjectRepository(CashDepositEntity)
+    private TDI17: Repository<CashDepositEntity>
+  ) {}
+
+  //TODO ADD DTO AND RETURN TYPE
+  async addSales(body: any[]) {
```
vyasworks, line 20: we wont need this for sales, it should pipe in through the API

vyasworks: if we are transporting data dumps, then it should live else where and not with production code.

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    return body.forEach(async (itm: any) => {
+      await this.salesRepo.save(this.salesRepo.create(itm));
+    });
+  }
+
+  //TODO ADD DTO
+  async addTDI34(body: any) {
```
vyasworks, line 27: I think the promises are not handled properly here. 

Using async in a loop, do we need to await on Promise.all?

vyasworks: if these are running sequentially then we need to make it parallel and use promise.all

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -0,0 +1,84 @@
...
+        })
+      );
+    });
+  }
+
+  //TODO ADD DTO AND RETURN TYPE
+  async addTDI17(body: any) {
+    body.details.forEach(async (itm: any) => {
+      return await this.TDI17.save(
+        this.TDI17.create({
```
vyasworks, line 51: create inside save?

docker-compose.yml
```diff
@@ -1,30 +1,57 @@
 version: "3.7"
 services:
+  db:
+    image: postgres:13-alpine
```
vyasworks, line 4: We can use PG v14 since RDS has support for it.

docker-compose.yml
```diff
@@ -1,30 +1,57 @@
...
     ports:
       - "3000:3000"
     environment:
-      - CHOKIDAR_USEPOLLING="true"
+      - DB_HOST=db
+      - DB_NAME=${DB_NAME}
+      - DB_USER=${DB_USER}
+      - DB_PASSWORD=${DB_PASSWORD}
+      - ALLOW_EMPTY_PASSWORD=yes
+      - CHOKIDAR_USEPOLLING=true
```
vyasworks, line 41: Add comment to explain what this is

vyasworks (changes requested):
few comments and qns to address, try not to do a force push since pr is a bit long :) 

- Migrations can be added in the next PR.

vyasworks:
NEXT PR 

- store the event as a jsonb as well for referencing in reconciliation outputs. 
- Add integration test cases that prove that additional fields are accepted and persisted in jsonb. 
- update the JSON spec for sales api to store in postgres / skip the sales stream and S3 stuff.

apps/backend/src/reconciliation/entities/static.entity.ts
chelsea-EYDS: yes for sure 👍🏻

apps/backend/src/reconciliation/reconciliation.service.ts
chelsea-EYDS: This is a shortened version of:

```
const entity = await this.entity.create({entityData})
if (entity){
  await this.entity.save(entity)
}
```

chelsea-EYDS: The repository pattern in typeorm requires a create and then save. I am not sure why, but I suppose if you wanted to create an entity and maybe perform some other logic on it, or pass it to another function before saving?

chelsea-EYDS:
To dos:
- address any TODO in the code - ie: proper param and return types
- write migrations, update database module
- review and enforce uniqueness for entities (this is currently disabled - restrictions were causing inserts to fail - maybe b/c of test data?)
- update naming to be consistent (ie sales -> transaction, tdi34 -> pos deposit, tdi17 -> cash deposit, etc)
- fix the location id/payment types in the parsing from TDI17 as per conversation yesterday with Karim
- map reconciliation components with enums
- store the event as a jsonb as well for referencing in reconciliation outputs.
- Add integration test cases that prove that additional fields are accepted and persisted in jsonb
- update the JSON spec for sales api to store in postgres / skip the sales stream and S3 stuff.

apps/backend/src/flat-files/tdi17/TDI17Details.ts
```diff
@@ -123,7 +123,7 @@ export class TDI17Details
     this.resource.location_desc = data;
   }
 
-  @Column({ start: 67, width: 12 })
+  @Column({ start: 67, width: 12, format: { type: DataType.Decimal } })
```
chelsea-EYDS, line 126: **note: I am not using float here as this requires further parsing
The string is passed in as something like: "00009900"
The last two digits are always the two decimal places, as described in the TDI definition files. We need to insert a decimal prior to parsing it as a float.

chelsea-EYDS:
> NEXT PR
> 
> * store the event as a jsonb as well for referencing in reconciliation outputs.
> * Add integration test cases that prove that additional fields are accepted and persisted in jsonb.
> * update the JSON spec for sales api to store in postgres / skip the sales stream and S3 stuff.

I have included these points in a list and Graeme is creating our tickets :)

chelsea-EYDS:
Merging as per conversation 01/06 4:30pm

PR #23, opened 2023-01-06
SFTP Transfer family and RDS postgres setup 

vyasworks:


terraform/sftp.tf
```diff
@@ -0,0 +1,112 @@
...
+            ],
+            "Resource": ["${aws_s3_bucket.sftp_storage.arn}/*"] 
+        }
+    ]
+}
+POLICY
+}
+
+resource "aws_s3_bucket" "sftp_storage" {
+  bucket = "pcc-integration-data-files-${var.target_env}"
```
chelsea-EYDS, line 66: I just wanted to double check that we're planning to use a different bucket name than the current one of: "bc-pcc-data-files-<env>" ?

chelsea-EYDS (approved):
I am still lacking context on a few points but it all looks good to me.

PR #24, opened 2023-01-10
Create payment-noplugins.yaml

eyjwarren:
Test API Gateway config generated from OpenAPI spec using gwa tool


chelsea-EYDS (commented):
This looks fine to me! I think Vyas should review as I don't have enough context on this one. One thing I am realizing though is the sales api dto needs to be updated to reflect the latest spec. I'm not sure if this is relevant to your PR, so if not, please disregard :)

vyasworks (approved):
LGTM. 

Please add a readme with documentation on what this file means, where / how it will or should be applied.

vyasworks:
@eyjwarren

PR #28, opened 2023-01-11
Update sales endpoint

chelsea-EYDS:
**What has changed:**

Temporary update to sales api to parse the data used in reconciliation and ignore the distributions for the time being.


apps/backend/src/sales/sales.service.ts
```diff
@@ -17,13 +18,13 @@ export class SalesService {
     try {
       await (
         await this.firehoseService.putRecord(
```
vyasworks, line 20: We can stop using firehose since we pivoted to DB.

apps/backend/src/sales/sales.service.ts
```diff
@@ -17,13 +18,13 @@ export class SalesService {
...
-          event as unknown as Record<string, string>
+          parseGarms(event) as unknown as Record<string, string>
```
vyasworks, line 21: The sales end point is generic, it should not need parseGarms()

apps/backend/src/sales/sales.service.ts
```diff
@@ -17,13 +18,13 @@ export class SalesService {
...
         )
       ).promise();
```
vyasworks, line 23: The event should go to the DB here, it can be the next PR though, we can send it to the void for now.

vyasworks (changes requested):

apps/backend/src/sales/sales.service.ts
chelsea-EYDS: Haha, no ideas why I decided call parseGarms here. I wanted to return the parsed garms as some sort of response, but I am doing that already outside of the try/catch. 

I intentionally did not include the db update here as we are currently ingesting the sales files over sftp, and if we started manually adding in test data here it might mess things up ?

vyasworks (approved):

PR #34, opened 2023-01-27
Local + AWS - S3 file parsing based on event types

vyasworks:
Add Pos Module
Add Cash Module
Local File Parsing
S3 event Notification Support
Delta file parse and saves
TDI cash, pos repositories
update makefiles


chelsea-EYDS (approved):

apps/backend/src/s3-manager/s3-manager.service.ts
```diff
@@ -16,20 +16,13 @@ export class S3ManagerService {
     return response.Contents?.map((c) => c.Key);
   }
 
-  async getContents(bucket: string, key: string) {
+  async getObject(bucket: string, key: string) {
```
chelsea-EYDS, line 19: We need to update this in the lambdas I think?

PR #37, opened 2023-01-28
Match And Kill Round 1

chelsea-EYDS:


apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
       throw err;
     }
   }
+
+  public async getMerchantIds(location_id: number): Promise<number[]> {
```
vyasworks, line 63: this should be `getMerchantIdsByLocationId()`

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+    const merchant_ids = await this.transactionRepo.manager.query(`
+        SELECT DISTINCT "Merchant ID"
```
vyasworks, line 65: This function should be in location service

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
       throw err;
     }
   }
+
+  public async getMerchantIds(location_id: number): Promise<number[]> {
+    const merchant_ids = await this.transactionRepo.manager.query(`
+        SELECT DISTINCT "Merchant ID" 
+        FROM public.master_location_data ml 
+        WHERE ml."GARMS Location" = ${location_id} 
+        AND "Type" != 'Bank'
```
vyasworks, line 68: add `Bank` in constants.ts

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+      `);
+    return merchant_ids.map((itm: any) => parseInt(itm['Merchant ID']));
+  }
+
+  public async getPTIDs(location_id: number): Promise<string[]> {
```
vyasworks, line 73: The naming is unclear here, should it be `getPTLocdByGarmsLocId`

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+    const pt_ids = await this.transactionRepo.manager.query(`
+          SELECT DISTINCT "Location"
```
vyasworks, line 75: Fn shd be in location service

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+      `);
+    return merchant_ids.map((itm: any) => parseInt(itm['Merchant ID']));
+  }
+
+  public async getPTIDs(location_id: number): Promise<string[]> {
+    const pt_ids = await this.transactionRepo.manager.query(`
+          SELECT DISTINCT "Location" 
+          FROM public.master_location_data ml 
+          WHERE ml."GARMS Location" = ${location_id}
+        `);
```
vyasworks, line 78: Location Repository should be able to query this without a RAW query?

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+    return pt_ids.map((itm: any) => itm['Location']);
+  }
+
+  public async getLocationList(): Promise<ILocation[]> {
```
vyasworks, line 82: Location service. 

We should not use `transactionRepo` for this

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+        ORDER BY "GARMS Location" DESC;
+      `);
+    return locations.map((itm: any) => ({
+      sbc_location: itm['GARMS Location'],
+      office_name: itm.description
+    }));
+  }
+
+  public async getMethod(sbc_code: number) {
+    const payment_method = await this.paymentMethodRepo.findOne({
```
vyasworks, line 96: Shd live in payment Service

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+      where: { sbc_code }
+    });
+    if (payment_method) {
+      return payment_method.sbc_code;
+    }
+  }
+
+  public async reconcilePOS(date: string, location_id: number) {
+    const locations = await this.getLocationList();
```
vyasworks, line 105: Semantics are confusing here, making it hard to read a bit. 

the parameter `location_id` -> denotes pos location or garms loc?

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+  public async getMethod(sbc_code: number) {
+    const payment_method = await this.paymentMethodRepo.findOne({
+      where: { sbc_code }
+    });
+    if (payment_method) {
+      return payment_method.sbc_code;
+    }
+  }
+
+  public async reconcilePOS(date: string, location_id: number) {
```
vyasworks, line 104: Should we name this `reconcilePOSBySalesLocation`? 

Also, this is a great example of what I call `Layer`

apps/backend/src/sales/sales.service.ts
```diff
@@ -38,4 +38,58 @@ export class SalesService {
       throw e;
     }
   }
+
+  async queryCashTransactions(
```
vyasworks, line 42: Unsure about the function of this query in this service, please clarify.

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -30,4 +59,146 @@ export class ReconciliationService {
...
+    const total_transactions_amt = pos_transactions?.reduce(
+      (a: any, b: any) => a + parseFloat(b.amount),
+      0
+    );
+
+    const total_deposit_amt = pos_deposits?.reduce(
+      (a: any, b: any) => a + parseFloat(b.transaction_amt),
+      0
+    );
+
```
vyasworks, line 134: We can add line by line matching here?

vyasworks (commented):
I understand this is a draft PR. 

I have added my comments (you might have already considered them or they may be obvious) 

I think this is in a good shape to work on top of too add more `layers` and `heuristics`

apps/backend/src/reconciliation/reconciliation.service.ts
```diff
@@ -1,32 +1,143 @@
...
+    private posService: PosService,
+    @Inject(CashService)
+    private cashService: CashService,
+    @Inject(LocationService)
+    private locationService: LocationService
+  ) {}
 
-  //TODO this is temporary for testing the parsed garms json only
-  async readAndParseGarms(filename: string, filebuffer: Buffer) {
+  // TODO define return type
```
chelsea-EYDS, line 26: Define Return Types/Outputs

Add another match layer

apps/backend/src/sales/sales.service.ts
chelsea-EYDS: This should return all of the payments for Cash/Cheque (to match against TDI17). 

I am trying to query by the current and last deposit dates. This will probably need to be changed. I have not yet found a rule for date range that works for each location. 

We should add a `and match=false`. Maybe then we can query for anything less than the current deposit date, as all of the previous payments should be accounted for? (and if they are not then an exception could be thrown?)

vyasworks (approved):

vyasworks:
WIP Merge... Follwup PR to fix issues

PR #38, opened 2023-01-31
Add NestJS to package.json/dependencies

eyjwarren:
Fix pcc-backend error "nest: command not found"

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,4 +26,8 @@
       "eslint  --fix"
     ]
   }
+  "dependencies": {
+    "@nestjs/common": "^9.2.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,4 +26,8 @@
       "eslint  --fix"
     ]
   }
+  "dependencies": {
+    "@nestjs/common": "^9.2.1",
```
chelsea-EYDS, line 30: Not sure if it matters, but should we remove the carat to use the specific version here?

vyasworks:
to use nest cli without installing it globally use `/payment-common-component $ ./node_modules/.bin/nest`

PR #43, opened 2023-02-03
[REFACTOR] Match & kill

chelsea-EYDS:


apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -26,7 +26,7 @@ export class POSReconciliationService {
       const payment = payments.find(
         (itm) =>
           itm.amount == parseFloat(deposit.transaction_amt.toString()) &&
-          itm.method == deposit.method
+          itm.method == deposit.card_vendor
```
chelsea-EYDS, line 29: This will not match, as the method is an sbc_garms code, and the pos deposit table is using a join to set method to match this code. This will require updates elsewhere.

chelsea-EYDS: (I know that as discussed we are planning to use the master data payment method table to match all payments/deposits, but at the time of writing this I was still under the assumption that we'd be handling all joins post insert to the db, and that reconciliation would be SBC specific for MVP). I can set the join on the payment table to match the card vendor if that is preferable?

apps/backend/.eslintrc.js
```diff
@@ -14,13 +13,14 @@ module.exports = {
     jest: true
   },
   rules: {
+    'no-console': 'off',
+    '@typescript-eslint/no-explicit-any': 'error',
```
vyasworks, line 17: we want to enforce and follow this going forward, avoid any unless there is an absolute need to have it. 

WIP and transitional work should still use types known at the time of the feature work.

apps/backend/src/deposits/entities/pos-deposit.entity.ts
```diff
@@ -43,12 +52,23 @@ export class POSDepositEntity {
   @Column({ default: false })
   match: boolean;
 
+  // rename this to just payment id?
   @Column({ nullable: true })
   matched_payment_id?: string;
 
-  method?: number;
+  @ManyToOne(() => PaymentMethodEntity, (pd) => pd.method)
```
vyasworks, line 59: We need these join columns to properly get the entities with the relationships, including lookups. 

This makes querying so much easier and readable.

apps/backend/src/deposits/entities/pos-deposit.entity.ts
```diff
@@ -43,12 +52,23 @@ export class POSDepositEntity {
...
+  @JoinColumn({ name: 'card_vendor' })
+  payment_method: PaymentMethodEntity;
 
   constructor(data?: TDI34Details) {
     Object.assign(this, data?.resource);
   }
+
+  public get timestamp(): Date {
```
vyasworks, line 67: use `getters` and `setters` to manipulate, cast, format data etc.. 

it is better to do it here than do casting in every logic where the property is used.

apps/backend/src/deposits/entities/pos-deposit.entity.ts
vyasworks: This is only a typeorm config and does not create a column

apps/backend/src/deposits/pos-deposit.service.ts
```diff
@@ -19,55 +19,36 @@ export class PosDepositService {
...
+      where: {
+        transaction_date: event?.date,
+        metadata: {
+          program: event?.program
+        },
+        merchant_id: In(merchant_ids)
+      },
+      // Order by needs to be in this order for matching logic. 
+      // We need to batch them using order to ease matches
+      order: {
```
vyasworks, line 44: this is ordering is necessary for EAGER match. so that the matches actually are working based on card vendor AND time

apps/backend/src/deposits/pos-deposit.service.ts
```diff
@@ -78,29 +59,25 @@ export class PosDepositService {
...
-  }
-
   async createPOSDepositEntity(
     data: POSDepositEntity
   ): Promise<POSDepositEntity> {
     return await this.posDepositRepo.save(this.posDepositRepo.create(data));
   }
 
-  async updatePOSDepositEntity(
-    deposit: Partial<POSDepositEntity>,
```
vyasworks, line 95: There should ideally be no use for `Partial types` in our work so far, all our entities are fully defined and loaded. 

This issue stems because sql queries are used in getting the data and creating the entity classes. 

This is unsafe because it will overwrite the data...if we save the entity directly (which we arent doing here)

apps/backend/src/reconciliation/cash-reconciliation.service.ts
```diff
@@ -19,60 +20,62 @@ export class CashReconciliationService {
...
+    deposits: CashDepositEntity[],
+    payments: PaymentEntity[]
+  ): CashPaymentsCashDepositPair | void {
+    const sumOfPayments = payments.reduce(
+      (sum, payment) => sum + payment.amount,
+      0
+    );
 
-  public async match(deposits: CashDepositEntity[], payments: PaymentEntity[]) {
-    const matched = deposits.map((itm: CashDepositEntity) => ({
```
vyasworks, line 39: There are several issues here: 

- typecasting in logic rather than class getters. 

- having partial entities restrict the amount of matching we can do here, this is still ok since we do only amts but will constrain us if the function has scope beyond it. 

- having DB queries in a loop same as that of the business logic. Although this is not a huge performance penalty now, it make it harder to process the logic. It also makes logging, auditing etc harder. 

- too many nested promises make debugging and control flow harder.

apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -19,58 +21,83 @@ export class POSReconciliationService {
...
+    const alreadyMatchedPosDeposits = deposits.filter((itm) => itm.match);
+    const yetToBeMatchedPosDeposits = deposits.filter((itm) => !itm.match);
+
+    const matched_in_this_run = this.matchPosPaymentToPosDeposits(
+      yetToBeMatchedPosPayments,
+      yetToBeMatchedPosDeposits
     );
 
-    const matched =
-      payments && deposits && (await this.match(deposits, payments));
```
vyasworks, line 68: This issue of having to validate the variable before using it won't occur if we use proper types on all parameters.

vyasworks: here payments and deposits are arrays, so I am not sure why we need to validate them, and if they are empty, match function should still be able to handle it,.

apps/backend/src/transaction/entities/payment.entity.ts
```diff
@@ -5,22 +5,27 @@ import {
...
 export class PaymentEntity {
   @PrimaryGeneratedColumn('uuid')
   id: string;
 
-  @Column({ type: 'numeric' })
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
+  // https://typeorm.io/entities#column-options - For better numeric representation.
+  // https://github.com/typeorm/typeorm/issues/873#issuecomment-424643086 - And make this column return proper number
+  // 16+4 should be standard for amounts in PCC
+  @Column({ type: 'numeric', precision: 16, scale: 4, transformer: new ColumnNumericTransformer(), })
```
vyasworks, line 22: Defining `amount` / `currency` types properly in the database will save us from LOT of issues. 

there will be no need for type casting the amounts anywhere, anymore. DB stores this with precision and proper scale of decimals. 

Accessing in type script is also in proper `number` format with help of `class-transformer`. No more hidden runtime numeric strings that need typecasting!

vyasworks: going forward every `currency` column should use this same definition or modified only based on business needs.

vyasworks (approved):

.github/workflows/jira.yaml
```diff
@@ -0,0 +1,25 @@
+name: Check PR For JIRA Issue
```
chelsea-EYDS, line 1: 👀

Makefile
```diff
@@ -229,6 +229,9 @@ reconcile:
 	@docker exec -it $(PROJECT)-backend ./node_modules/.bin/ts-node -e 'require("./apps/backend/src/lambdas/reconcile.ts")'
 
 report:
+	@docker exec -it $(PROJECT)-backend ./node_modules/.bin/ts-node -e 'require("./apps/backend/src/lambdas/report.ts").handler()'
```
chelsea-EYDS, line 232: 👀

apps/backend/src/transaction/transaction.service.ts
```diff
@@ -57,95 +60,82 @@ export class TransactionService {
...
+        transaction: {
+          fiscal_close_date: event.date,
+          location_id: event.location_id
+        }
+      }
+    });
   }
 
+  // TODO: move this to payments service
+  // TODO: we need the entities, so cannot work with raw queries!
```
chelsea-EYDS, line 84: 👀

apps/backend/src/lambdas/utils/parseGarms.ts
```diff
@@ -27,14 +27,13 @@ export const parseGarms = async (
       new TransactionEntity({
         source_id: Ministries.SBC,
         id: sales_transaction_id,
-        transaction_date: sales_transaction_date
+        transaction_date: sales_transaction_date // TODO: just use slice here
```
vyasworks, line 30: TODO: Re review parse garms spec, save the json, accounting, etc..

vyasworks: TODO: Test cases for garms mapping

apps/backend/src/lambdas/utils/parseGarms.ts
```diff
@@ -27,14 +27,13 @@ export const parseGarms = async (
...
           .split('')
           .splice(0, 10)
           .join(''),
         transaction_time: sales_transaction_date
-          .split('')
-          .splice(11, 10)
-          .join(''),
+          .slice(11, 19)
```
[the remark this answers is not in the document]
vyasworks: TODO: Move this logic to the entity as `parseGarmsToTxn` @chelsea-EYDS

apps/backend/src/transaction/entities/payment.entity.ts
vyasworks: @chelsea-EYDS - Review and update `amount column` make a Typeorm Decorator for the same so it's even across

PR #49, opened 2023-02-08
[CHORE] Fix heuristic bug on eager matching

vyasworks:
Objective: 

Fix bug that does not mark `match=true` for certain deposits when they were matched. 

Issue - reuse of deposits for matching after they were already used for matches!

### DIFF
--- a/apps/backend/src/reconciliation/pos-reconciliation.service.ts
+++ b/apps/backend/src/reconciliation/pos-reconciliation.service.ts
@@ -32,31 +32,34 @@ export class POSReconciliationService {
 
     // Basic EAGER! 1:1 matching
 
-    for (const payment of payments) {
+    for (const [pindex, payment] of payments.entries()) {
       // console.log(`Processing payment for ${payment.amount} - ${payment.id} - ${payment.timestamp}`)
-      for (const deposit of deposits) {
+      for (const [dindex, deposit] of deposits.entries()) {
         // console.log(`Processing deposit ${deposit.id} for ${deposit.transaction_amt} - ${deposit.timestamp}`)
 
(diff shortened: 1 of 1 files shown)


apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -32,32 +32,35 @@ export class POSReconciliationService {
...
 
         // TODO:make this a strategy pattern
         if (
           payment.amount === deposit.transaction_amt &&
           payment.method === deposit.card_vendor &&
+          deposit.match === false &&
           differenceInSeconds(payment.timestamp, deposit.timestamp) < 240
         ) {
+
+          // mutate the original array to use in next heurisitc match
```
vyasworks, line 48: This similar problem will exist in cash totals as well, needs fix

chelsea-EYDS: Could we implement the layered approach here, but invert it. So first we check 1:1 matching where difference in seconds is very small, then if no matches, continue to broaden the window until there are none yet left to be matched for the date?

chelsea-EYDS: Okay, so there is no case in which a transaction and deposit could be delayed greater than 240 seconds?

chelsea-EYDS (approved):

apps/backend/src/reconciliation/pos-reconciliation.service.ts
vyasworks: there could be but having a greater time span make more possibility for bad matches than good ones? this could be wrong but how I understand it now. we can verify with BAs.

PR #50, opened 2023-02-09
[CCFPCM-349] Logger

chelsea-EYDS:
[CCFPCM-349](https://bcdevex.atlassian.net/browse/CCFPCM-349)

**Objective:** 

Create a new transport in the current AppLogger Service to log to a file. 

Enable optional database logging. 

**What has changed:**
- Most files are import updates for the appLogger
- Added a custom logger which implements typeorm logger
- updated existing logger to use new transport
- optional typeorm logging to view all raw sql queries generated. To enable add: process.env.DB_LOGS_ENABLED=true to .env


apps/backend/src/database/database.module.ts
```diff
@@ -11,6 +12,14 @@ import { TypeOrmModule } from '@nestjs/typeorm';
         username: process.env.DB_USERNAME ?? 'postgres',
         password: process.env.DB_PASWORD ?? 'postgres',
         database: process.env.DB_NAME ?? 'pcc',
+        logging:
+          process.env.DB_LOGS_ENABLED === 'true'
```
chelsea-EYDS, line 16: Update conditional

eyjwarren (approved):
Reviewed & built

apps/backend/src/database/helpers.ts
```diff
@@ -0,0 +1,8 @@
+import { LoggerOptions } from 'typeorm';
+import { DatabaseLogger } from './database-logger';
+
+export const loggerOptions: LoggerOptions = ['query', 'error', 'schema'];
+export const logLevels =
+  process.env.DB_LOGS_ENABLED === 'true' ? loggerOptions : undefined;
```
vyasworks, line 6: why undefined?

apps/backend/src/database/datasource.ts
```diff
@@ -24,5 +27,7 @@ export default new DataSource({
     LocationEntity
   ],
   migrations: [__dirname + '/migrations/**/*{.ts,.js}'],
-  synchronize: false
+  synchronize: false,
+  logging: logLevels,
+  logger: dbLogger
```
vyasworks, line 32: could be conditional insert using spread operator

apps/backend/src/logger/entities/log.entity.ts
```diff
@@ -0,0 +1,24 @@
+import {
```
vyasworks, line 1: to be removed, will include when we find a use case.

apps/backend/src/logger/logger.service.ts
```diff
@@ -1,22 +1,49 @@
...
+    @InjectRepository(LogEntity) private logsRepository: Repository<LogEntity>
+  ) {
     this.logger = WinstonModule.createLogger({
-      transports: [new winston.transports.Console({})],
+      transports: [
+        new winston.transports.Console({
+          level: 'info',
+          format: consoleLogFormat
+        }),
+        new DailyRotateFile({
```
vyasworks, line 25: TODO: make this optional based on the environment

vyasworks (changes requested):

apps/backend/src/database/helpers.ts
chelsea-EYDS: The Base Datasource logger options does not include false as a type, but since the logger itself is optional we can intentionally set this to undefined in the datasource config.

apps/backend/src/database/datasource.ts
chelsea-EYDS: I'm not sure what you mean here

chelsea-EYDS: (I know what a spread operator and a conditional is, I'm just unclear on how you want them used here)

apps/backend/src/logger/entities/log.entity.ts
chelsea-EYDS: done :)

apps/backend/src/logger/logger.service.ts
chelsea-EYDS: done :)

apps/backend/src/database/datasource.ts
eyjwarren: @vyasworks : Are you referring to another place in the code? I don't see where the spread operator could be used in this definition, so I'm going to approve this PR and we can fix it later if there's a better way to do it.

eyjwarren (approved):

vyasworks (approved):

apps/backend/src/database/datasource.ts
vyasworks: yes, I think my PR on running migrations will include it. so good for now.

PR #52, opened 2023-02-11
[CHORE] Add request-transformer to payment-noplugins.yaml

eyjwarren:
Aidan Cope: "In our test environment we have a global request-transformer plugin that adjusts the Host to align with what would be in production if preserve_host=true. This will only create an issue in our test instance, not our prod instance. To get around this add a 'request-transformer' plugin to your Service - so that it takes precedence over the global one."


vyasworks:
Needs ticket number.

PR #63, opened 2023-02-28
[CCFPCM-0369] pos-amounts

chelsea-EYDS:
[CCFPCM-0369](https://bcdevex.atlassian.net/browse/CCFPCM-0369)

Objective: 

Parse TDI34 refunds
Script to update current db entries


vyasworks (changes requested):
This should be done WHILE PARSING and not as a post parsing DB operation.

Makefile
```diff
@@ -338,4 +338,11 @@ open-db-tunnel:
...
 	aws ec2-instance-connect send-ssh-public-key --instance-id $(BASTION_INSTANCE_ID) --instance-os-user ec2-user --ssh-public-key file://ssh-keypair.pub
-	ssh -i ssh-keypair ec2-user@$(BASTION_INSTANCE_ID) -L 5454:$(DB_HOST):5432 -o ProxyCommand="aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
\ No newline at end of file
+	ssh -i ssh-keypair ec2-user@$(BASTION_INSTANCE_ID) -L 5454:$(DB_HOST):5432 -o ProxyCommand="aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
+
+# =====================================
+# Database Scripts
+# =====================================
+
+adjust-pos-refunds:
```
vyasworks, line 347: To be removed.

apps/backend/src/deposits/pos-deposit.service.ts
```diff
@@ -92,4 +93,22 @@ export class PosDepositService {
     depositEntity.status = MatchStatus.EXCEPTION;
     return await this.posDepositRepo.save(depositEntity);
   }
+
+  async adjustAmountsForRefundTransactions() {
```
vyasworks, line 97: This logic should be in the parser.

vyasworks (changes requested):

apps/backend/src/deposits/pos-deposit.service.ts
chelsea-EYDS: This is just for the current data that is already in the DB, since we've already deployed our database

chelsea-EYDS:
> This should be done WHILE PARSING and not as a post parsing DB operation.

That is correct - this change has been included in the TDI 34 Details class. 

The script for updating the old data could run once and then be removed

vyasworks:
> > This should be done WHILE PARSING and not as a post parsing DB operation.
> 
> That is correct - this change has been included in the TDI 34 Details class.
> 
> The script for updating the old data could run once and then be removed (unless there is a better way to update the old data entries?)

Correct, we should write migrations for that. (port the controller function to a new migration file) 

it is not feasible to run make commands for each of these envs, manually.

chelsea-EYDS:
> > > This should be done WHILE PARSING and not as a post parsing DB operation.
> > 
> > 
> > That is correct - this change has been included in the TDI 34 Details class.
> > The script for updating the old data could run once and then be removed (unless there is a better way to update the old data entries?)
> 
> Correct, we should write migrations for that. (port the controller function to a new migration file)
> 
> it is not feasible to run make commands for each of these envs, manually.

Okay sounds good. I think I have resolved this, please re-review when you have time.

apps/backend/src/common/utils/formatFixedWidth.ts
```diff
@@ -1,26 +1,3 @@
-export const transactionCode = (code: string) => {
```
vyasworks, line 1: Reason for removal? 
Was this used before?

vyasworks (approved):

apps/backend/src/common/utils/formatFixedWidth.ts
chelsea-EYDS: When we first built out the parsing I had included this as a nested field, but then removed in order to keep the flat files "flat" as we had discussed handling any mapping etc after the flat files had been parsed. (This was pre-database, I think we were still using the files in cyber duck at that time ?)

chelsea-EYDS: There was no criteria or specs re: the adjustment to the transaction_amt field based on these values at that time.

PR #64, opened 2023-03-01
[CCFPCM-0393] non CAD cash

chelsea-EYDS:
[CCFPCM-0393](https://bcdevex.atlassian.net/browse/CCFPCM-0393)

**Objective:** 

- Account for USD during reconciliation

- Add column to record the calculated CAD amount in the case of non CAD payments (based on the provided exchange rate)

- Write migrations to include this column with the proper amounts in the db

- Write migration to update the current garms SBC exchange rate format to a standard exchange rate format. (Percentage, from foreign currency to CAD, not vice versa... ie: USD -> CAD = 1.34)

- Update garms parsing for proper exchange rate format

- Update API spec to validate proper exchange rate format

- Update cash reconciliation to use the calculated amount

- Add a getter to the payment class to return adjusted amount if it is exists, or, amount if not, and use this value for cash reconciliation

### DIFF
--- a/apps/backend/sample-files/transaction.json
+++ b/apps/backend/sample-files/transaction.json
@@ -17,7 +17,7 @@
     "payments": [
       {
         "amount": 20.0,
-        "forgein_currency_amount": 15.0,
+        "foreign_currency_amount": 15.0,
         "currency": "USD",
         "exchange_rate": 1.34,
         "payment_method": "CASH",
--- a/apps/backend/sample-files/transaction.json
+++ b/apps/backend/sample-files/transaction.json
@@ -4,7 +4,7 @@
     "transaction_date": "2023-01-01",
     "transaction_time": "13.33.31.875973",
     "fiscal_close_date": "2023-01-02",
-    " total_transaction_amount": 25.2,
+    "total_transaction_amount": 30.2,
     "void_indicator": false,
     "miscellaneous": {
       "employee_id": "SC61350"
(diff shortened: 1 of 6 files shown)


apps/backend/src/transaction/dto/payment.dto.ts
```diff
@@ -50,15 +56,29 @@ export class PaymentDTO {
...
   amount!: number;
 
+  @ApiProperty({
+    description: 'Total CAD amount - Equal to amount paid * exchange rate',
+    example: 134.0
+  })
+  @ApiProperty()
+  @IsNumber()
+  @IsOptional()
+  adjusted_amount?: number;
```
v-y-a-s, line 66: this should be renamed to `cad_amount` 

it cannot be optional

apps/backend/src/transaction/entities/payment.entity.ts
```diff
@@ -117,6 +126,10 @@ export class PaymentEntity {
     Object.assign(this, payment);
   }
 
+  public get payment_amount(): number {
+    return this.adjusted_amount ?? this.amount;
```
v-y-a-s, line 130: I dont think this is right. both these fields could be of different currencies

apps/backend/sample-files/transaction.json
```diff
@@ -12,20 +12,20 @@
...
       {
-        "amount": 10.2,
-        "currency": "CAD",
-        "exchange_rate": 0,
-        "payment_method": "CASH",
+        "amount": 15.0,
+        "adjusted_amount": 20.12,
+        "currency": "USD",
+        "exchange_rate": 1.34,
+        "payment_method": "US_FUNDS",
```
v-y-a-s, line 23: is this a new payment method because it exists in GARMS? 

do we need `US_FUNDS`? then is CAD payments `CAD_FUNDS`? 

if we say just CASH,  is it CAD or any currency cash?

vyasworks (approved):

PR #65, opened 2023-03-02
[FIX] filtering zero amount in parse garms

chelsea-EYDS:

Objective: 

- Filter out zero dollar amount when parsing garms SBC files
- Update cash payments query
- Write migration to remove current zero dollar amount payments from the db

### DIFF
--- a/apps/backend/src/database/migrations/1677775665682-migration.ts
+++ b/apps/backend/src/database/migrations/1677775665682-migration.ts
@@ -1,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migration1677775665682 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+        DELETE FROM PAYMENT WHERE AMOUNT = 0::numeric;
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+        DELETE FROM PAYMENT WHERE AMOUNT = 0::numeric;
+    `);
+  }
+}
--- a/apps/backend/src/lambdas/utils/parseGarms.ts
+++ b/apps/backend/src/lambdas/utils/parseGarms.ts
@@ -17,60 +17,66 @@ export const parseGarms = async (
   source_file_name: string,
   paymentMethods: PaymentMethodEntity[]
 ): Promise<TransactionEntity[]> => {
-  return garmsJson.map(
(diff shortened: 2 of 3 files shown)


apps/backend/src/database/migrations/1677775665682-migration.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migration1677775665682 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+        DELETE FROM PAYMENT WHERE AMOUNT = 0::numeric;
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
```
chelsea-EYDS, line 10: Not sure that this is being done properly here... ?

vyasworks: do we have to delete? this will make database short of entries than what is in the files.

apps/backend/src/lambdas/utils/parseGarms.ts
```diff
@@ -17,60 +17,66 @@ export const parseGarms = async (
...
-          .join(''),
-        transaction_time: sales_transaction_date
-          .slice(11, 19)
-          .replaceAll('.', ':'),
-        location_id: parseInt(source.location_id),
-        total_transaction_amount: payment_total,
+  return garmsJson
+    .map((itm) => ({
+      ...itm,
+      payments: itm.payments.filter((payment) => payment.amount !== 0)
```
vyasworks, line 23: did we find out why certain amounts are 0 in the first place? 

what is the reasoning behind removal?

vyasworks:
Just had a chat with Karim abt this, he wants to remove these entries from garms so ✔️  for me.

vyasworks (approved):

apps/backend/src/database/migrations/1677775665682-migration.ts
chelsea-EYDS: As per discussion with Karim this morning, he suggested that we filter out the zero dollar payments from garms as we should (hopefully) not be getting these in the future (since zero dollar amounts cannot be reconciled as there is not corresponding deposit). 

I have not updated the api spec to validate this as he mentioned we should discuss with the team on whether or not to reject incoming $0.00 payments through the api, or whether we should continue to filter. (I think it would be best to reject a $0.00 rather than filter once we get this data through the api...?)

PR #66, opened 2023-03-02
[CCFPCM-0404] TDI 17 location id -> pt_location_id

chelsea-EYDS:

Objective: 

TDI 17 “transaction type“ and “location_id“ should be a single column/field called pt_location_id. 

Update this in the TDI17 parser, update cash match location lookup, and write migrations to fix this issue.

### DIFF
--- a/apps/backend/src/deposits/cash-deposit.service.ts
+++ b/apps/backend/src/deposits/cash-deposit.service.ts
@@ -53,12 +53,12 @@ export class CashDepositService {
     const {
       program,
       date,
-      location: { location_id }
+      location: { pt_location_id }
     } = event;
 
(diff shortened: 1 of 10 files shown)


apps/backend/src/deposits/cash-deposit.service.ts
```diff
@@ -53,12 +53,12 @@ export class CashDepositService {
     const {
       program,
       date,
-      location: { location_id }
+      location: { pt_location_id }
```
vyasworks, line 56: where is `pt_location_id` calculated? 

when the input is given as just `location_id` in any event, how does this work?

vyasworks (approved):

PR #68, opened 2023-03-02
[FIX] drop match column pos_deposit

chelsea-EYDS:

**Objective:** 

- remove "match" column from pos_deposit


eyjwarren (approved):
Reasonable to move the default to SQL

PR #69, opened 2023-03-02
[FIX] cash location_ids

chelsea-EYDS:

**Objective:** 

- When the column for the pt_location_ids in the cash deposit table was last updated, the logic for the location lookup for cash reconciliation had an error if a location id was passed in with the reconciliation event. If a location_id is passed in to the recon event, this location_id should be used to lookup pt_location_ids, rather than expecting a pt_location_id to be passed in with the event.

### DIFF
--- a/apps/backend/fixtures/lambda/reconcile.json
+++ b/apps/backend/fixtures/lambda/reconcile.json
@@ -1,6 +1,6 @@
 {
   "fiscal_start_date": "2023-01-01",
-  "fiscal_close_date": "2023-02-27",
+  "fiscal_close_date": "2023-03-02",
   "program": "SBC",
   "location_ids": []
 }
--- a/apps/backend/src/lambdas/reconcile.ts
+++ b/apps/backend/src/lambdas/reconcile.ts
@@ -37,10 +37,14 @@ export const handler = async (
     event.location_ids.length === 0
       ? await locationService.getLocationsBySource(event.program)
       : await locationService.getLocationsByID(event);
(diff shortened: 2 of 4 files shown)


apps/backend/fixtures/lambda/reconcile.json
```diff
@@ -1,6 +1,6 @@
 {
   "fiscal_start_date": "2023-01-01",
-  "fiscal_close_date": "2023-02-27",
+  "fiscal_close_date": "2023-03-02",
```
eyjwarren, line 3: This range is just for testing purposes, right? What's the best way to mark it so that it gets changed to the real fiscal year before this is being used in production?

chelsea-EYDS: I think we could remove the dates from the event and just use an environment variable to set the date. If the plan is to run reconciliation daily starting from the start of the fiscal year, or quarter, etc, then we shouldn't need to pass these in. 
We also probably don't need the option to pass in locations or programs either. It makes sense to look up by location for reporting, but if this is meant to run daily for all locations and programs then we can set these with env vars... I think we weren't clear on the criteria so it was left as flexible as possible for the time being.

chelsea-EYDS: (I think this is something we should clarify with the team though and could be handled in another PR as this one will unblock testing).

eyjwarren (approved):
TO DO: figure out how to handle fiscal dates

PR #73, opened 2023-03-10
[CHORE] clean up matching for cash - improve logs

chelsea-EYDS:
[CHORE](https://bcdevex.atlassian.net/browse/CCFPCM-0000)

**Objective:** 

Clean up the current cash match. Match only one deposit date per date per location. 

Improve logging.

New output (Deposit ID can be used to query the matching payments for testing): 

Screenshot 2023-03-10 at 3 12 52 PM

### DIFF
--- a/apps/backend/src/deposits/cash-deposit.service.ts
+++ b/apps/backend/src/deposits/cash-deposit.service.ts
@@ -1,6 +1,6 @@
 import { Inject, Injectable, Logger } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-import { Repository, LessThanOrEqual } from 'typeorm';
+import { Raw, Repository, LessThanOrEqual } from 'typeorm';
 import { CashDepositEntity } from './entities/cash-deposit.entity';
 import { MatchStatus } from '../common/const';
 import { AppLogger } from '../logger/logger.service';
@@ -60,11 +60,11 @@ export class CashDepositService {
       where: {
         pt_location_id,
         metadata: { program: program },
-        deposit_date: LessThanOrEqual(date),
+        deposit_date: date,
         status: status
       },
       order: {
-        deposit_date: 'ASC',
+        deposit_date: 'DESC',
         deposit_amt_cdn: 'DESC'
       }
     });
@@ -72,13 +72,13 @@ export class CashDepositService {
 
   public async findPastDueDate(
     event: ReconciliationEvent
-  ): Promise<{ currentDate: string; pastDueDate: string | null }> {
+  ): Promise<{ currentDate: string; pastDueDate: string } | void> {
     const {
       date,
       program,
       location: { pt_location_id }
     } = event;
-
+    if (!date) return;
     const dates = await this.cashDepositRepo.find({
       select: { deposit_date: true },
       where: {
@@ -93,11 +93,8 @@ export class CashDepositService {
     const distinctDepositDates = Array.from(
       new Set(dates.map((item) => item.deposit_date))
     );
-    if (distinctDepositDates.length < 4)
-      return {
-        currentDate: date,
-        pastDueDate: null
-      };
+    if (distinctDepositDates.length < 4) return;
+
     return {
       currentDate: date,
       pastDueDate: distinctDepositDates[3]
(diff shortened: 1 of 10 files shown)


apps/backend/src/location/location.service.ts
```diff
@@ -31,39 +31,6 @@ export class LocationService {
     return merchant_ids.map((itm) => itm.merchant_id) as number[];
   }
 
-  public async getPTLocationsByID(
```
chelsea-EYDS, line 34: Removed - we can get pt_location_id from the same query used in POS reconciliation

apps/backend/src/deposits/cash-deposit.service.ts
```diff
@@ -60,25 +60,25 @@ export class CashDepositService {
       where: {
         pt_location_id,
         metadata: { program: program },
-        deposit_date: LessThanOrEqual(date),
+        deposit_date: date,
         status: status
       },
       order: {
-        deposit_date: 'ASC',
+        deposit_date: 'DESC',
```
chelsea-EYDS, line 67: Keeping consistency with the order. Cash Match will still run in ascending order from the specified fiscal start -> fiscal_close, but the queries for payments and deposits will be ordered in descending order from the deposit date which is nearest (less than or equal) to the current reconciliation event date

apps/backend/src/deposits/cash-deposit.service.ts
```diff
@@ -93,11 +93,8 @@ export class CashDepositService {
     const distinctDepositDates = Array.from(
       new Set(dates.map((item) => item.deposit_date))
     );
-    if (distinctDepositDates.length < 4)
-      return {
-        currentDate: date,
-        pastDueDate: null
-      };
+    if (distinctDepositDates.length < 4) return;
```
chelsea-EYDS, line 96: Don't throw exceptions if we have less than 4 deposit dates of data to look back to

chelsea-EYDS: (should we make this 3?)

apps/backend/src/deposits/cash-deposit.service.ts
```diff
@@ -113,27 +110,50 @@ export class CashDepositService {
...
+        pt_location_id: location.pt_location_id,
         metadata: { program },
-        deposit_date: LessThanOrEqual(date)
+        deposit_date: Raw(
+          (alias) =>
+            `${alias} <= :fiscal_close_date and ${alias} > :fiscal_start_date `,
+          {
+            fiscal_close_date,
+            fiscal_start_date
+          }
```
chelsea-EYDS, line 126: Get the list of dates between the fiscal_start and fiscal_close. I can't see a need to match outside of the specified date range.

apps/backend/src/deposits/cash-deposit.service.ts
```diff
@@ -113,27 +110,50 @@ export class CashDepositService {
...
+    this.appLogger.log(
+      `UPDATED: ${deposits.length} CASH DEPOSITS to ${status.toUpperCase()}`,
+      CashDepositService.name
+    );
 
+    return await Promise.all(
+      deposits.map(
+        async (deposit) =>
+          await this.updateDepositStatus({ ...deposit, status })
+      )
```
chelsea-EYDS, line 154: I've moved the mapping for status updates out of cash reconciliation service in order to clean it up a bit.

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -38,12 +38,6 @@ export const handler = async (
       ? await locationService.getLocationsBySource(event.program)
       : await locationService.getLocationsByID(event);
   appLogger.log(`Found ${locations.length} pos_deposit locations`);
-  const pt_locations =
-    event.location_ids.length === 0
-      ? await locationService.getPTLocationsBySource(event.program)
-      : await locationService.getPTLocationsByID(event);
-
```
chelsea-EYDS, line 45: Not needed - can use the same as pos and specify pt_location_id in the queries

apps/backend/src/transaction/payment.service.ts
```diff
@@ -78,12 +78,12 @@ export class PaymentService {
         status,
         transaction: {
           location_id,
-          fiscal_close_date: LessThanOrEqual(date)
+          fiscal_close_date: LessThanOrEqual(date!)
         }
       },
```
chelsea-EYDS, line 83: TO DO: fiscal_close_date less than or equal to date (current deposit date) and greater than or equal to (previous deposit date? or, fiscal_start_date in the reconciliation event?)

apps/backend/src/transaction/payment.service.ts
```diff
@@ -129,13 +129,31 @@ export class PaymentService {
...
 
+  async updatePayments(
+    payments: PaymentEntity[],
+    status: MatchStatus
+  ): Promise<PaymentEntity[]> {
+    this.appLogger.log(
+      `UPDATED: ${payments.length} PAYMENTS to ${status.toUpperCase()}`,
+      PaymentService.name
+    );
+
```
chelsea-EYDS, line 140: Moved into payment service to clean up cash recon service

PR #74, opened 2023-03-13
[CCFPCM-376] Excel Export Base

vyasworks:
[CCFPCM-376](https://bcdevex.atlassian.net/browse/CCFPCM-376)

Objective:


apps/backend/src/app.module.ts
```diff
@@ -26,6 +27,7 @@ import { TransactionModule } from './transaction/transaction.module';
     ParseModule,
     LocationModule,
     ExceptionModule,
+    ExcelexportModule,
```
chelsea-EYDS, line 30: ExcelExport?

chelsea-EYDS (approved):
Looks good. Could we rename the Excelexport to 'ExcelExport' ? I am happy to make this change with the work for the first reporting sheet i f you agree on the rename?

vyasworks:
> Looks good. Could we rename the Excelexport to 'ExcelExport' ? I am happy to make this change with the work for the first reporting sheet i f you agree on the rename?

sure sounds good, pls rename when you pull in !

PR #78, opened 2023-03-20
[CCFPCM-417] implement map limit for parsing

vyasworks:
[CCFPCM-417](https://bcdevex.atlassian.net/browse/CCFPCM-417)

### DIFF
--- a/apps/backend/src/common/promises.ts
+++ b/apps/backend/src/common/promises.ts
@@ -1,0 +1,23 @@
+export async function mapLimit<T, R>(
+  arr: T[],
+  fn: (x: T , i: number) => Promise<R>,
+  limit = 10
+): Promise<R[]> {
+  if (limit < 1) {
+    throw new Error('limit must be >= 1');
+  }
+  const results: R[] = [];
+  await Promise.all(
+    arr.slice(0, limit).map(async (x, i) => {
+      results[i] = await fn(x, i);
+      while (true) {
+        const next = limit++;
+        if (next >= arr.length) {
+          break;
+        }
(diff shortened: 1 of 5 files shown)


apps/backend/src/deposits/cash-deposit.service.ts
```diff
@@ -34,12 +35,16 @@ export class CashDepositService {
...
+  async saveCashDepositEntities(
+    data: CashDepositEntity[]
+  ): Promise<CashDepositEntity[]> {
     try {
-      return await this.cashDepositRepo.save(this.cashDepositRepo.create(data));
-    } catch (err) {
-      this.appLogger.error(err);
-      throw err;
+      const entities = data.map((d) => this.cashDepositRepo.create(d));
+      return mapLimit(entities, (entity) => this.cashDepositRepo.save(entity));
```
chelsea-EYDS, line 44: I think we could also use forEach here and not return anything (we're not using the returned values). I believe forEach is about 30% faster?

apps/backend/src/common/promises.ts
```diff
@@ -0,0 +1,23 @@
+export async function mapLimit<T, R>(
+  arr: T[],
+  fn: (x: T , i: number) => Promise<R>,
+  limit = 10
+): Promise<R[]> {
```
chelsea-EYDS, line 5: nice

apps/backend/src/deposits/cash-deposit.service.ts
vyasworks: we would still need the map limit to throttle calls to db.

PR #80, opened 2023-03-22
[FIX] pos-deposit-match

chelsea-EYDS:

**Objective:** 

- pos_deposit id was not being recorded. 

- added in second layer of matching to match by day, after matching by 5 minutes initially for POS.

- cleaned up the reconcile lambda logging

- (attempted) to simplify some of the pos reconciliation functions for easier unit testing

- removed unused or unnecessary functions

### DIFF
--- a/apps/backend/src/reconciliation/cash-reconciliation.service.ts
+++ b/apps/backend/src/reconciliation/cash-reconciliation.service.ts
@@ -59,6 +59,13 @@ export class CashReconciliationService {
       dateRange,
       location
     );
+    if (depositDates.length < 2) {
+      this.appLogger.log(
+        'No past due dates found',
+        CashReconciliationService.name
+      );
+      return;
+    }
     const dates = {
       currentDate: depositDates[0],
       pastDueDate: depositDates[2]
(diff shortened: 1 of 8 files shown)


apps/backend/src/reconciliation/cash-reconciliation.service.ts
```diff
@@ -58,8 +61,7 @@ export class CashReconciliationService {
     );
     const dates = {
       currentDate: depositDates[0],
-      previousDate: depositDates[2],
-      pastDueDate: depositDates[3]
+      pastDueDate: depositDates[2]
```
vyasworks, line 64: check for array length so that this does not become undefined.

apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -1,252 +1,268 @@
...
-      matchedDeposits: await this.posDepositService.markPosDepositsAsMatched(
-        matches.filter((itm) => itm.deposit)
-      )
-    };
+  ): Promise<void> {
+    this.appLogger.log(
+      `MATCHES: ${matches.length}`,
+      POSReconciliationService.name
+    );
+    matches.forEach(async (itm) => {
```
vyasworks, line 141: await promise.all?

vyasworks (commented):
left some comments

vyasworks (approved):

PR #84, opened 2023-04-05
CCFPCM395 mock data and some unit tests

chelsea-EYDS:
[CCFPCM-0395](https://bcdevex.atlassian.net/browse/CCFPCM-0395)

Objective: 
- Create mock data generator for POS Deposits, Payments and Cash Deposits
- Integration Test for file parsing
- basic set up for unit testing deposit modules
- WIP/TODO reconciliation testing (base set up started)


apps/backend/src/database/database.module.ts
```diff
@@ -25,12 +23,26 @@ const getEnvironmentSpecificConfig = (env?: string) => {
...
       return {
+        synchronize: false,
+        migrationsRun: false,
         entities: [join(__dirname, '../**/*.entity.js')],
         migrations: [join(__dirname, '../migration/*.js')],
         logging: ['migration'] as LoggerOptions
       };
+    case 'test':
+      return {
+        synchronize: true,
```
vyasworks, line 34: sync: true - reasoning?

vyasworks (approved):

apps/backend/src/database/database.module.ts
chelsea-EYDS: Intention is to recreate a temporary testing db in the container on each run in the pipeline so I thought we'd need to set sync true in order to not run the migrations. Please let me know if there is a better way to configure this?

PR #88, opened 2023-04-13
[CHORE] Date types 

chelsea-EYDS:

Objective:
**3 PRs in one, sorry it is huge, merged to test them all together before merging into main**

- Updated master location data
- Migrations for master location data changes
- Updated payment method master data
- Migrations for payment method master data
- Updated the logic for querying payments in order to match to pos_deposit vs cash_deposit table
- Updated joins on pos_deposit and payment entities to map to the payment_method tab;e
- Migrations for the changes above
- Cleaned up parse garms code slightly
- Small report code refactor
- Changes/fixes for the cash report details
- Dates: attempted to update codebase to use type Date but ran into some issues, see notes below

**notes:**
- Dates:
	- Postgres 'date' type is a string
	- In our application we have been using date in some place and string in others - I attempted to make the dates consistent by using the typescript type "Date" throughout the application
	- issues:
		- Date is a timestamp: (mm-DD-yyyy HH:MM:ss)
		- The dates passed in from our lambdas and during parsing are strings, formated as dates: 'yyyy-MM-dd'. They do not contain a time component. 
		- The timestamp seems better suited for a generated date column such as a createdAt or updatedAt timestamp, rather than a date string without the time
		- We should not use a typescript Date type in our class attributes if we are using a postgres 'date' column (even though this will work fine on insert) because when we query using find({where:{date}) the database 'date' column value will be in the format of 'yyyy-MM-dd' whereas the passed in find operator value will be a timestamp (mm-DD-yyyy HH:MM:ss) and will not match (although there will be type error as the class definition is using Date)
		
		
- Migrations and master data:
	- When we update the master data lists, we need to write migrations to reflect these changes. Even though the master data is the last migration to run, it will not in our deployed environments because it has already ran, and we have data in this env (vs dropping your schema locally and then running the migrations to create the database schema)
- I have created a new ticket to handle the master data via lambda functions rather than through the migrations

- Payment Method
	- Previously we had been querying for payment using  the payment "method" as method Not(In(...pos payment methods)) 
	- We need a way to differentiate between the payments to match to pos_deposit vs cash_deposit that does not require prior knowledge of which payment method maps to which bank deposit file
	- I have started by adding a new classification column to the payment methods and then querying the payment method table for the relevant types but we may want to refine/revisit this for easier maintainability as new clients onboard


apps/backend/src/deposits/entities/cash-deposit.entity.ts
```diff
@@ -25,7 +25,7 @@ export class CashDepositEntity {
   program_code?: string;
 
   @Column({ type: 'date' })
-  deposit_date: string;
+  deposit_date: Date | string;
```
vyasworks, line 28: I think this should be just `Date` and then be casted to string of typeorm needs it to be that way with a pre db hook or something similar.

chelsea-EYDS: I might need your help on this. I've tried using "@BeforeInsert" hook from typeorm but I'm getting the same error :/

chelsea-EYDS: I suppose we could switch to using a "timestamp" in the db vs "date"?

chelsea-EYDS: This article explains the typeorm issue:

https://wanago.io/2021/03/15/postgresql-typeorm-date-time/

apps/backend/src/database/migrations/1681495109117-migration.ts
```diff
@@ -0,0 +1,71 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migration1681495109117 implements MigrationInterface {
+  name = 'migration1681495109117';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
```
vyasworks, line 7: Have all of these alter columns been tested with data? 

is the casting of data working by default for existing data? Has it been validated?

apps/backend/src/database/migrations/1681495109117-migration.ts
```diff
@@ -0,0 +1,71 @@
...
+      `ALTER TABLE "transaction" ALTER COLUMN "transaction_time" TYPE varchar(8))`
+    );
+    await queryRunner.query(
+      `ALTER TABLE "transaction" ALTER COLUMN "fiscal_close_date" TYPE DATE`
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "pos_deposit" ALTER COLUMN "transaction_date" TYPE DATE`
+    );
+    await queryRunner.query(
```
vyasworks, line 53: I am unsure if data will auto cast for this?

vyasworks (commented):
I am ok to approve once the migrations queries are clarified, 

I think this PR needs an qa right after deployment to test on dates stuff specifically.

apps/backend/src/database/migrations/1681495109117-migration.ts
chelsea-EYDS: Updated: using ::type 
Tested with existing db data

apps/backend/src/database/migrations/1679690933840-migration.ts
```diff
@@ -1,15 +0,0 @@
-import { MigrationInterface, QueryRunner } from 'typeorm';
```
vyasworks, line 1: why are the migrations deleted? have they not run already in dev?

vyasworks (commented):
Good to merge pending migration qns..

vyasworks (approved):

PR #89, opened 2023-04-17
[CCFPCM-458] master_data table updates

chelsea-EYDS:
[CCFPCM-458](https://bcdevex.atlassian.net/browse/CHORE)

Objective: 
- update master data tables
- small clean up for readability in parse garms function
- removed method from payment table and updated code to use the joined payment_method
- added filetype in payment_method for easier distinction between TDI17 and TDI34 payment types 
 
payment table ER diagram:

Screenshot 2023-04-17 at 4 02 15 PM


apps/backend/src/transaction/entities/payment.entity.ts
```diff
@@ -86,7 +83,7 @@ export class PaymentEntity {
   // PCC - Internals
 
   @ManyToOne(() => PaymentMethodEntity, (pm) => pm.method)
-  @JoinColumn({ name: 'method' })
+  @JoinColumn([{ name: 'payment_method', referencedColumnName: 'method' }])
```
chelsea-EYDS, line 86: Seemed redundant to have a 'method' as well as a 'payment_method'. Using the joined column for payment_method throughout the codebase. Using the payment_method.deposit_file_type to ease querying for cash payment types (using method not in ('AX', 'M', 'V', 'P') is less reliable since we have more TDI34 payment methods than in this list)

vyasworks: I am not sure I understand. Both the tables now have the column `method` right?

apps/backend/src/transaction/entities/payment-method.entity.ts
```diff
@@ -10,6 +12,9 @@ export class PaymentMethodEntity {
   @Column('varchar', { length: 2, nullable: true })
   sbc_code?: string;
 
+  @Column({ enum: FileTypes, nullable: true })
```
vyasworks, line 15: payment method master is agnostic of the file types. This is a master defined by PCC. I am not sure if we need the file types here...

apps/backend/src/transaction/payment.service.ts
```diff
@@ -31,12 +31,14 @@ export class PaymentService {
           location_id: location.location_id
         },
         status: paymentStatus,
-        method: In(pos_payment_methods)
+        payment_method: {
```
vyasworks, line 34: This is a wrong approach, linking two unrelated domain models. 

TDI 34 has nothing to do with payments looking objectively from a transactions point of view. This creates unnecessary coupling.

vyasworks (changes requested):

apps/backend/src/transaction/payment.service.ts
chelsea-EYDS: Resolved as per conversation

PR #96, opened 2023-05-02
[CHORE] split cash exceptions

chelsea-EYDS:

Objective: 
 - improve readability by splitting out the exceptions for cash into a separate service
 - easier unit testing


apps/backend/src/lambdas/reconcile.ts
```diff
@@ -170,14 +174,13 @@ const runCashExceptions = async (
...
         : event.period.from;
-      console.log(pastDueDate);
       if (pastDueDate) {
         appLogger.log(
           `Processing CASH EXCEPTIONS for: ${location.description} (${location.location_id})`,
           CashReconciliationService.name
         );
         appLogger.log(
-          `EXCEPTIONS DATE: ${pastDueDate}`,
+          `EXCEPTIONS DATE: ${cashDates[dindex - 2]}`,
```
fw-noel, line 183: I could use an explanation on what `cashDates` looks like. I'm curious why we choose to log this particular value rather than pastDueDate

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -151,7 +154,8 @@ const runCashExceptions = async (
   event: ReconciliationConfigInput,
   locations: LocationEntity[],
   appLogger: Logger,
-  cashRecon: CashReconciliationService
+  cashRecon: CashReconciliationService,
+  exceptionsService: CashExceptionsService
```
fw-noel, line 158: Argument for renaming to `cashExceptionsService` for consistency?

apps/backend/src/reconciliation/cash-reconciliation.service.ts
```diff
@@ -157,11 +137,11 @@ export class CashReconciliationService {
         location
       );
     this.appLogger.log(
-      `${aggregatedPayments.length} AGGREGATED PAYMENTS PENDING RECONCILIATION`,
+      `${aggregatedPayments?.length} AGGREGATED PAYMENTS PENDING RECONCILIATION`,
```
fw-noel, line 140: If this nullish is necessary, then I imagine you'd want `aggregatedPayments?.length || 0`. Otherwise, the log won't look right!

chelsea-EYDS: Oops, good call! On this topic, I was wondering if I should scale back the logs? Or maybe add more logging to the specific services that I are querying for/updating the data vs having the logs here? The logs are bit unwieldily...

fw-noel: Logs are cheap, no need to scale back, but we may want to refine them so as to make them easier to sift through.

apps/backend/src/lambdas/reconcile.ts
fw-noel: Explained!

fw-noel (approved):

apps/backend/src/lambdas/reconcile.ts
chelsea-EYDS: I'm not sure why this is not showing as outdated? It has been changed :)

chelsea-EYDS: Screenshot 2023-05-09 at 8 09 41 PM

chelsea-EYDS:
Merging, approved by Noel, but when the other PR merged into this one it reverted the approval.

fw-noel (approved):

PR #101, opened 2023-05-04
[CCFPCM-419] pos reconciliation service and pos deposit service unit tests

chelsea-EYDS:
[CCFPCM-0419](https://bcdevex.atlassian.net/browse/CCFPCM-0419)

Objective:

- POS reconciliation service unit tests
- POS deposit unit tests

### DIFF
--- a/apps/backend/test/unit/reconciliation/helpers.ts
+++ b/apps/backend/test/unit/reconciliation/helpers.ts
@@ -1,0 +1,78 @@
+import { faker } from '@faker-js/faker';
+import { differenceInMinutes, format, getTime, parse } from 'date-fns';
+import { POSDepositEntity } from '../../../src/deposits/entities/pos-deposit.entity';
+import { PaymentEntity } from '../../../src/transaction/entities';
+import { paymentMethods } from '../../../test/mocks/const/payment-methods';
+import { locations } from '../../mocks/const/locations';
+import { MockData } from '../../mocks/mocks';
+
+export const setSomePaymentsToTwentyMinutesLater = (
+  payments: PaymentEntity[]
+) => {
+  return payments.map((itm: PaymentEntity, index: number) =>
+    index % 2 === 0
+      ? {
+          ...itm,
+          timestamp: itm.timestamp,
+          transaction: {
+            ...itm.transaction,
+            transaction_time: setTransactionTimeAheadBy20Minutes(
+              itm.transaction.transaction_date,
+              itm.transaction.transaction_time
+            )
+          }
+        }
+      : {
+          ...itm,
+          timestamp: itm.timestamp
+        }
+  );
+};
+
+export const setTransactionTimeAheadBy20Minutes = (
+  date: string,
+  time: string
+) =>
+  format(
+    getTime(parse(`${date} ${time}`, 'yyyy-MM-dd HH:mm:ss', new Date())) +
+      1000 * 60 * 20,
+    'HH:mm:ss'
+  );
+
+export const unmatchedTestData = (
+  data: MockData
+): { payments: PaymentEntity[]; deposits: POSDepositEntity[] } => {
+  const paymentData = data.paymentsMock;
+  const posData = data.depositsMock as POSDepositEntity[];
+  return {
+    payments: paymentData.map((itm) => ({
+      ...itm,
+      transaction: {
+        ...itm.transaction,
+        transaction_date: format(faker.date.recent(10), 'yyyy-MM-dd'),
+        location: faker.helpers.arrayElement(locations)
+      },
+      timestamp: parse(
+        format(faker.date.recent(10), 'yyyy-MM-dd'),
+        'yyyy-MM-dd',
+        new Date()
+      ),
+      amount: faker.datatype.number({ min: 1, max: 1000 }),
+      payment_method: faker.helpers.arrayElement(paymentMethods)
+    })),
+    deposits: posData.map((itm) => ({
+      ...itm,
+      timestamp: itm.timestamp,
+      transaction_amt: faker.datatype.number({ min: 1, max: 1000 })
+    }))
+  };
+};
+
+export const timeBetweenMatchedPaymentAndDeposit = (
+  payment: PaymentEntity,
+  deposit: POSDepositEntity
+) =>
+  differenceInMinutes(
+    parse(payment.transaction.transaction_time, 'HH:mm:ss', new Date()),
+    parse(deposit.transaction_time, 'HH:mm:ss', new Date())
+  );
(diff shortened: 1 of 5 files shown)


apps/backend/test/unit/reconciliation/pos-reconciliation.spec.ts
```diff
@@ -1,58 +1,311 @@
...
+    const mockPOSData = new MockData(PaymentMethodClassification.POS);
+
+    payments = mockPOSData.paymentsMock;
+    /**
+     * Set the transaction time to be 20 minutes after the transaction date for one payment
+     */
+    payments[0].transaction.transaction_time = format(
+      getTime(
+        parse(
+          payments[0].transaction.transaction_date +
```
fw-noel, line 44: Can we use template literals \`${}\` here?

Also, an explanation for the 20 minutes thing? Is it just for the one test?

apps/backend/test/unit/reconciliation/pos-reconciliation.spec.ts
```diff
@@ -1,58 +1,311 @@
...
+    expect(logger).toBeDefined();
+  });
+
+  it('should verify the test data transaction_time', () => {
+    expect(
+      differenceInMinutes(
+        parse(payments[0].transaction.transaction_time, 'HH:mm:ss', new Date()),
+        parse(deposits[0].transaction_time, 'HH:mm:ss', new Date())
+      )
+    ).toBeGreaterThanOrEqual(20);
```
fw-noel, line 79: What exactly are we testing here? Just that the 20 minute addition worked? It feels like we're just testing date-fns `differenceInMinutes`

apps/backend/test/unit/reconciliation/pos-reconciliation.spec.ts
```diff
@@ -1,58 +1,311 @@
...
+    expect(deposits.every((itm) => itm.status === MatchStatus.PENDING)).toBe(
+      true
+    );
+  });
+
+  describe('matchPaymentsToDeposits heurisitics round one', () => {
+    it('should match payments to deposits according to first round of heuristics', () => {
+      const spy = jest.spyOn(service, 'matchPosPaymentToPosDeposits');
+      const roundOneTimeDiff = 5;
+      const timeDiff = roundOneTimeDiff;
```
fw-noel, line 95: Point of `roundOneTimeDiff`?

apps/backend/test/unit/reconciliation/pos-reconciliation.spec.ts
```diff
@@ -1,58 +1,311 @@
...
+      /**
+       * Call function to match based on first round of heuristics
+       */
+      matchedRoundOne = service.matchPosPaymentToPosDeposits(
+        payments,
+        deposits,
+        timeDiff
+      );
+      matches = [...matchedRoundOne];
+      expect(spy).toBeCalled();
```
fw-noel, line 106: Is there a scenario that this test fails?

apps/backend/test/unit/reconciliation/pos-reconciliation.spec.ts
```diff
@@ -1,58 +1,311 @@
...
+      /**
+       * Call function to match based on the second round of heuristics
+       */
+      matchedRoundTwo = service.matchPosPaymentToPosDeposits(
+        payments,
+        deposits,
+        timeDiff
+      );
+      matches.push(...matchedRoundTwo);
+      expect(spy).toBeCalled();
```
fw-noel, line 148: I'm still wondering if there's a scenario in which this fails. And if so, how this is different from the other `matchedRoundOne` test, since it's the same function. I might need to read more on how the heuristics are working

apps/backend/test/unit/reconciliation/pos-reconciliation.spec.ts
```diff
@@ -1,58 +1,302 @@
...
+        ).toBeLessThanOrEqual(5)
+      );
+    });
+
+    it('should have updated the status from PENDING to MATCH', () => {
+      matchedRoundOne.forEach((itm) => {
+        expect(itm.deposit.status === MatchStatus.MATCH);
+        expect(itm.payment.status === MatchStatus.MATCH);
+      });
+    });
```
fw-noel, line 113: Do we know exactly how many matches we have, and how many non-matches we have? Is there an argument for adding that as a test too?

apps/backend/test/unit/reconciliation/helpers.ts
```diff
@@ -0,0 +1,79 @@
+import { faker } from '@faker-js/faker';
+import { differenceInMinutes, format, getTime, parse } from 'date-fns';
+import { MatchStatus } from '../../../src/common/const';
+import { POSDepositEntity } from '../../../src/deposits/entities/pos-deposit.entity';
+import { PaymentEntity } from '../../../src/transaction/entities';
+import { paymentMethods } from '../../../test/mocks/const/payment-methods';
+import { locations } from '../../mocks/const/locations';
+import { MockData } from '../../mocks/mocks';
+
+export const setSomePaymentsToTwentyMinutesLater = (
```
fw-noel, line 10: This is also "resetting" the payments right (as in setting the status to pending)?

I'd suggest just passing an array into this function and calling it something like `resetPaymentsDataset` or something, and doing the 20 minute thing for each even index within the function rather than passing in an index. With some comments to explain why.

chelsea-EYDS: The payments are reset between tests using:
(this is a recent change :) )

```
  afterEach(() => {
    jest.clearAllMocks();
  });
```

Function has been updated to receive an array, and status updates are removed (it was _redundant_ since the mock data generates the status as PENDING and each test will generate new data)

:)

chelsea-EYDS: sorry for the caps-lock on PENDING, I wasn't yelling... that's how it is written in the codebase (as an enum)

apps/backend/invoke.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
fw-noel, line 1: Probably should be in a separate pr

apps/backend/test/unit/reconciliation/helpers.ts
```diff
@@ -0,0 +1,73 @@
...
+            ...itm.transaction,
+            transaction_time: setTransactionTimeAheadBy20Minutes(
+              itm.transaction.transaction_date,
+              itm.transaction.transaction_time
+            )
+          }
+        }
+      : {
+          ...itm,
+          timestamp: itm.timestamp
```
fw-noel, line 27: Isn't `itm.timestamp` covered within `...itm`?

apps/backend/invoke.sh
chelsea-EYDS: oops wasn't planning to commit this one

fw-noel (approved):

PR #102, opened 2023-05-10
 [CCFPCM-422] cash reconciliation and cash deposit unit tests

chelsea-EYDS:
[CCFPCM-0422](https://bcdevex.atlassian.net/browse/CCFPCM-0422)

Objective: 

- unit tests for cash reconciliation svc


apps/backend/test/mocks/mocks.cash.spec.ts
```diff
@@ -31,4 +32,25 @@ describe('Tests the generated mock data', () => {
...
+        amount: itm.deposit_amt_cdn
+      })),
+      payments: aggregatedPayments(cashPaymentsMock).map((itm) => ({
+        date: itm.fiscal_close_date,
+        amount: itm.amount
+      }))
+    };
+    expected.deposits.forEach((deposit) =>
+      expected.payments.forEach((payment) =>
+        expect(
```
fw-noel, line 49: So I don't love this test, and initially I thought it was because it wasn't obvious what the expected value is supposed to look like, but ultimately, it's supposed to just be a bunch of 1s. So I think that's fine.

But I think what makes it less readable / maybe what I don't like about it is that it's unclear how many expects it is gonna go through. The double array iteration means it's running x*y times, which could be a lot, or could be 1 or 4 or 9.

I'm not sure what the solution is though. Could it be pulling some random indexes and comparing those? Or pulling some specific ones?

I guess my other question is, do we HAVE to test, e.g. deposits[2] with payments[0], payments[1], payments[2], payments[3] ...? That's currently what's happening right? Or do we just want to test deposits[2] with just payments[2]?

apps/backend/test/unit/deposits/cash-deposit.service.spec.ts
```diff
@@ -1,78 +1,68 @@
...
+            find: jest.fn(),
+            save: jest.fn()
           }
-        }
+        },
+        { provide: Logger, useValue: {} }
       ]
     }).compile();
 
-    service = module.get<CashDepositService>(CashDepositService);
```
fw-noel, line 46: I believe in your other PR, you kept the service variable name as `service`. Happy for consistency either way; I did kind of like the idea of naming it just `service`, as if to say "this is the service we're actually testing with this spec"

apps/backend/test/unit/deposits/cash-deposit.service.spec.ts
```diff
@@ -86,119 +76,229 @@ describe('CashDepositService', () => {
...
+        .spyOn(cashDepositRepo, 'find')
+        .mockResolvedValueOnce(expectedEntities as any);
+
+      const foundEntities = await cashDepositService.findCashDepositExceptions(
+        depositDate,
+        program,
+        location
+      );
+
+      expect(foundEntities).toEqual(expectedEntities);
```
fw-noel, line 276: I like the idea behind this test, but I'm not sure there's a failure case. `find` literally just returns specific values, so the cases in `findCashDepositExceptions` (like the status needing to be In Progress) isn't actually tested. I think ultimately, it's the difficult part of relying on mocks for tests.

Need to think about / discuss this.

apps/backend/test/mocks/mocks.cash.spec.ts
chelsea-EYDS: This one is just meant to test that the mock data is accurate - this particular test on line 47 is only meant to verify that the payments mock fiscal close date is earlier than the cash deposit_date, since deposits always occur after payments.

chelsea-EYDS: If the first date in the comparison (the deposit date) is before the second date in the comparison (the payment date), then the output from calling 'compareAsc' is -1, and the test will fail. If the dates are equal, then the output from calling 'compareAsc' is 0, and the test will fail. If the first date(the deposit date) is after the second date(the payment date) then calling 'compareAsc' will return with a 1, and the test will pass.

...probably a better way to check this...

chelsea-EYDS: (I deleted my initial response after a re-read. I am going to a refactor on this....all payment dates should be earlier than all deposit dates, so I think there's an easier way to do this.)

apps/backend/test/mocks/mocks.cash.spec.ts
```diff
@@ -31,4 +32,21 @@ describe('Tests the generated mock data', () => {
...
+    const depositsDates = cashDepositsMock.map((itm) =>
+      parse(itm.deposit_date, 'yyyy-MM-dd', new Date())
+    );
+    const paymentsDates = aggregatedPayments(cashPaymentsMock).map((itm) =>
+      parse(itm.fiscal_close_date, 'yyyy-MM-dd', new Date())
+    );
+    // compareAsc returns -1 if the first date is before the second date, which causes this function to fail
+    expect(
+      depositsDates.every((depositDate) =>
+        paymentsDates.every(
```
chelsea-EYDS, line 46: note: max number of deposits/payments is 30 and max number of deposits is 10 for the generated data

apps/backend/test/unit/deposits/cash-deposit.service.spec.ts
chelsea-EYDS: yes, I agree this test is not very useful. I think the cash-deposit service and pos-deposit service would be better tested as part of an e2e testing, since they're primarily just a pass through service to connect to the database. 

I'm definitely open to trying to create some more robust unit tests here, but there isn't a lot of logic to test as most of it is handled by typeorm... I'm definitely open to any ideas/suggestions on creating better tests for these services though!

PR #106, opened 2023-05-15
[CHORE] add trailing commas to prettier and run it

fw-noel:
No ticket attached, but is that okay?

Objective: 
I really wanted to add trailing commas so here it is


chelsea-EYDS:
I think you just need to format the PR title and message to follow the template, otherwise approved!

chelsea-EYDS (approved):

PR #109, opened 2023-05-17
[CHORE] refine logging process to fully utilize winston

fw-noel:
Note: this is pending review from devs if we like this logging style better / find it appropriate

Objective: 
- Logging to fully utilize the logging service that was already written up
- Logs to have a timestamp with timezone
- Ability to set context in the app logger to avoid having to specify it in every log message

Attaching screenshots for the difference in console

Before
Screenshot 2023-05-16 at 11 59 27 PM

After
Screenshot 2023-05-16 at 11 56 43 PM

### DIFF
--- a/apps/backend/src/lambdas/reconcile.ts
+++ b/apps/backend/src/lambdas/reconcile.ts
@@ -1,4 +1,4 @@
-import { Logger } from '@nestjs/common';
+import { AppLogger } from '../logger/logger.service';
 import { NestFactory } from '@nestjs/core';
 import { Context } from 'aws-lambda';
 import { eachDayOfInterval, format } from 'date-fns';
@@ -21,7 +21,7 @@ export const handler = async (
   const cashExceptionsService = app.get(CashExceptionsService);
   const locationService = app.get(LocationService);
   const reportingService = app.get(ReportingService);
-  const appLogger = new Logger();
+  const appLogger = new AppLogger('Reconciliation Handler');
 
   appLogger.log({ event });
   appLogger.log({ context });
@@ -41,7 +41,7 @@ export const handler = async (
     posReconciliationService
   );
 
-  appLogger.log({ posReconciliationRun }, POSReconciliationService.name);
+  appLogger.log({ posReconciliationRun });
 
   const cashReconciliationRun = await runCashReconciliation(
     event,
@@ -50,7 +50,7 @@ export const handler = async (
     cashReconciliationService
   );
 
-  appLogger.log({ cashReconciliationRun }, CashReconciliationService.name);
+  appLogger.log({ cashReconciliationRun });
 
   const cashExceptionsRun = await runCashExceptions(
     event,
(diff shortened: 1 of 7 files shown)


apps/backend/tsconfig.json
```diff
@@ -5,5 +5,5 @@
     "baseUrl": "./",
     "allowJs": true
   },
-  "exclude": ["node_modules", "dist", "**/*.spec.ts", "./env.js"]
+  "exclude": ["node_modules", "**/*.spec.ts", "./env.js", "./dist"]
```
chelsea-EYDS, line 8: I'm not sure if this would be relevant for this PR, but can we use absolute imports? My autoimport always imports 'src/ ...etc' which doesn't show up as an error until I run the lambda handler locally

apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -13,10 +13,12 @@ import { PaymentService } from '../transaction/payment.service';
...
 export class POSReconciliationService {
   constructor(
-    @Inject(Logger) private readonly appLogger: AppLogger,
     @Inject(PosDepositService) private posDepositService: PosDepositService,
-    @Inject(PaymentService) private paymentService: PaymentService
-  ) {}
+    @Inject(PaymentService) private paymentService: PaymentService,
+    private readonly appLogger: AppLogger
+  ) {
+    this.appLogger.setContext(`POSReconciliationService`);
```
chelsea-EYDS, line 20: 🆒

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -130,14 +129,10 @@ const runCashReconciliation = async (
       appLogger.log({ dateRange }, CashReconciliationService.name);
 
       appLogger.log(
-        `Processing Cash Reconciliation for: ${location.description} (${location.location_id}) for ${dateRange.from_date} - ${dateRange.to_date}`,
-        CashReconciliationService.name
+        `Processing Cash Reconciliation for: ${location.description} (${location.location_id}) for ${dateRange.from_date} - ${dateRange.to_date}`
```
chelsea-EYDS, line 132: I will try to avoid the erratic use of caps lock in the future. Maybe we can have a standard format for how to write the logs?

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -130,14 +129,10 @@ const runCashReconciliation = async (
...
       );
 
-      appLogger.log(
-        `CURRENT DEPOSIT DATE: ${dateRange.to_date}`,
-        CashReconciliationService.name
-      );
+      appLogger.log(`CURRENT DEPOSIT DATE: ${dateRange.to_date}`);
```
chelsea-EYDS, line 135: Just commenting on this one as a reference to my last comment

fwkendall:
Just checking status here - is this just waiting for an opportunity to address Chelsea's comments, or does it need another set of eyes or a ticket to get priority to wrap it up?

PR #116, opened 2023-05-29
[CCFPCM-486] Performance Updates for Reconciliation

fw-noel:
[CCFPCM-486](https://bcdevex.atlassian.net/browse/CCFPCM-486)

Objective: Optimize POS Reconciliation
- Reduce number of database requests, pull out retrieval and updates for every location rather than every date
- Use mapper to avoid every.every. We use a map of deposits[amount][date] to find if there are matching deposits for each payment
- Update Status instead of save. We also wrap it in a transaction. This should be much more performant than using save, [this](https://dev.to/rishit/optimizing-typeorm-tips-from-experience-part-1-dont-use-save-4ke9) is probably the simple explanation

Did not touch Cash Deposits yet, this could be made more performant for sure


apps/backend/src/deposits/pos-deposit.service.ts
```diff
@@ -22,18 +22,18 @@ export class PosDepositService {
...
     dateRange: DateRange,
     program: Ministries,
-    location_id: number,
+    location_ids: number[],
     statuses?: MatchStatus[]
   ): Promise<POSDepositEntity[]> {
     const depositStatuses = statuses ? statuses : MatchStatusAll;
     const { minDate, maxDate } = dateRange;
     const locations: LocationEntity[] =
       await this.locationService.getLocationsByID(
```
chelsea-EYDS, line 34: I think we can pass the merchant ids in from the reconcile lambda directly, since we're including those in the original location lookup? Just to save from making this extra query. (I might be wrong though, still reviewing :)

apps/backend/src/deposits/pos-deposit.service.ts
```diff
@@ -160,4 +166,34 @@ export class PosDepositService {
...
+   * This will update necessary deposits to a desired MatchStatus.
+   * This should be more performant as it is wrapped in a transaction
+   * @param deposits A list of deposit entities to set a new status for, with the expected heuristic round
+   * @returns {void}
+   */
+  async updateDepositStatus(
+    deposits: POSDepositEntity[],
+    status: MatchStatus
+  ): Promise<void> {
+    return this.posDepositRepo.manager.transaction(async (manager) => {
```
chelsea-EYDS, line 181: Nice, this is great. 

Can we return the UpdateResult type from typeorm instead of void? (Since we're calling return here.) I just think it will be more log-able/test-able this way.

chelsea-EYDS: ...also, I think we don't need to pass the status since we're setting the status earlier during the match process? Either we should pass in all the new values (ie status, heuristic round), or, destructure them all from the entity itself? (unless there is a reason for doing it this way that I am missing?)

apps/backend/src/deposits/pos-deposit.service.ts
```diff
@@ -86,7 +92,7 @@ export class PosDepositService {
     );
     return await this.posDepositRepo.find({
       where: {
-        transaction_date: LessThan(date),
+        transaction_date: LessThan(maxDate),
```
chelsea-EYDS, line 95: Can we rename this maxDate to something like "posExceptionsDate", or "pastDueDate", or something along those lines? I think min/max make sense when we're using a range. It would be good to set the variable for posExceptionsDate (or any other name that makes sense) in the lambda or reconciliation service in case we need to make adjustments later on. ?

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -32,33 +39,66 @@ export const handler = async (event: ReconciliationConfigInput) => {
           LocationMethod.Bank
         );
 
-  const cashDepositService = app.get(CashDepositService);
-  const dates: Date[] = eachDayOfInterval({
-    start: parse(event.period.from, 'yyyy-MM-dd', new Date()),
-    end: parse(event.period.to, 'yyyy-MM-dd', new Date()),
-  });
+  const masterLocations = await locationService.getLocationsByID(
```
chelsea-EYDS, line 42: I think we can use one call to the location service to get this result. If we remove the "LocationMethod" enum (which is confusing anyways, lol, my bad!) and get all locations regardless of the method, then we can call the reduce function on that first call, rather than querying twice for locations?

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -19,9 +22,13 @@ export const handler = async (event: ReconciliationConfigInput) => {
   const cashExceptionsService = app.get(CashExceptionsService);
   const cashReconciliationService = app.get(CashReconciliationService);
   const posReconciliationService = app.get(PosReconciliationService);
+  const cashDepositService = app.get(CashDepositService);
+  const paymentService = app.get(PaymentService);
+  const posDepositService = app.get(PosDepositService);
   const locationService = app.get(LocationService);
   const reportingService = app.get(ReportingService);
   const appLogger = app.get(Logger);
+  const dateRange = { minDate: event.period.from, maxDate: event.period.to };
```
chelsea-EYDS, line 31: Maybe we can add a check to say, "if it is todays date, then subtract one business day" to ensure we're not reconciling data on the current date?

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -32,33 +39,66 @@ export const handler = async (event: ReconciliationConfigInput) => {
...
+  );
 
   for (const location of locations) {
-    for (const date of dates) {
-      const reconciled = await posReconciliationService.reconcile(
-        location,
-        event.program,
-        date
-      );
+    const locationPayments = allPosPaymentsInDates.filter(
```
chelsea-EYDS, line 71: nice, way better

apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -42,32 +48,9 @@ export class PosReconciliationService {
...
-    );
-
-    const pendingDeposits = await this.posDepositService.findPOSDeposits(
-      dateRange,
-      program,
-      location.location_id,
-      [MatchStatus.PENDING, MatchStatus.IN_PROGRESS]
-    );
-
     if (pendingPayments.length === 0 && pendingDeposits.length === 0) {
```
chelsea-EYDS, line 54: I never know if this should be "&&" or "||" just adding this as a comment to remind myself, it is not part of your PR so please disregard, lol.

apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -84,85 +67,71 @@ export class PosReconciliationService {
...
-      ),
-      PosHeuristicRound.ONE
-    );
-
-    this.appLogger.log(
-      `MATCHES - ROUND ONE ${roundOneMatches.length}`,
-      PosReconciliationService.name
+    // Put deposits into a dictionary, first layer keys are amount, second layer keys are dates
+    // This makes it more efficient to find than looping through all deposits
+    const locationPosDepositsDictionary = pendingDeposits.reduce(
```
chelsea-EYDS, line 72: Should we be grouping these first by date, and then by amount? We want to make sure we're getting 1:1 matching for the first two heuristic rounds. (I might be missing something still reviewing)

apps/backend/src/deposits/pos-deposit.service.ts
fw-noel: To be done in a later PR

apps/backend/src/deposits/pos-deposit.service.ts
fw-noel: Will address both

apps/backend/src/deposits/pos-deposit.service.ts
fw-noel: Doing this

apps/backend/src/lambdas/reconcile.ts
fw-noel: Will not do

chelsea-EYDS:
excellleeeeent

chelsea-EYDS (approved):

PR #122, opened 2023-06-03
[CHORE]: reduce number of queries for location table

chelsea-EYDS:

Objective: 

- use "reduce" to create a nested location interface to use during reconciliation and reporting (instead of querying the location service in each of the deposit services by the location_id)


apps/backend/src/location/location.service.ts
```diff
@@ -14,36 +13,61 @@ export class LocationService {
...
-        method: locationMethod,
         location_id: In(location_ids),
       },
       order: {
         location_id: 'ASC',
       },
     });
+    return this.normalizeLocations(locations);
+  }
+  public normalizeLocations(locations: LocationEntity[]): NormalizedLocation[] {
```
fw-noel, line 29: Comment block

apps/backend/src/location/location.service.ts
```diff
@@ -14,36 +13,61 @@ export class LocationService {
...
+            resp_code: itm.resp_code,
+            service_line_code: itm.service_line_code,
+            stob_code: itm.stob_code,
+            project_code: itm.project_code,
+            merchant_ids: [],
+            pt_location_id: 0,
+            description: '',
+          };
+        }
+        itm.merchant_id !== 999999999 &&
```
fw-noel, line 48: I know why because you explained it to me, but this should be commented on why. We may also want to define `999999999` as a constant. We always want to avoid magic numbers

apps/backend/src/location/location.service.ts
```diff
@@ -14,36 +13,61 @@ export class LocationService {
...
+          acc[key].merchant_ids.push(itm.merchant_id);
+        if (itm.method === 'Bank') {
```
fw-noel, line 50: use `LocationMethod` enum since it exists?

fw-noel (commented):
I like this PR a lot! Please address comments

apps/backend/src/location/location.service.ts
chelsea-EYDS: done (not showing here as outdated but can be seen in diff...)
Screenshot 2023-06-06 at 7 08 05 PM

fw-noel (approved):

PR #123, opened 2023-06-03
[CCFPCM-0498] Report updates (post rounds 3/4 heuristics)

chelsea-EYDS:
[CCFPCM-0498](https://bcdevex.atlassian.net/browse/CCFPCM-0498)

Objective:


apps/backend/src/lambdas/report.ts
```diff
@@ -1,16 +1,268 @@
...
+  );
+  const matchedDeposits = matchedDepositsAndExceptions.filter(
+    (itm: POSDepositEntity) => itm.status === MatchStatus.MATCH
+  );
+  const depositExceptions = matchedDepositsAndExceptions.filter(
+    (itm: POSDepositEntity) => itm.status === MatchStatus.EXCEPTION
+  );
+  const matchedDepositIdsFromPayments =
+    matchedPayments.map((itm: PaymentEntity) => itm.pos_deposit_match) ?? [];
+  /*eslint-disable */
```
fw-noel, line 193: Why?

apps/backend/src/lambdas/report.ts
```diff
@@ -1,16 +1,268 @@
...
+  const matchedDepositsFromAnotherDay: POSDepositEntity[] = [];
+
+  matchedDepositIdsFromPayments.forEach((itm: POSDepositEntity | undefined) => {
```
fw-noel, line 196: You could filter `matchedDepositIdsFromPayments` beforehand to avoid `| undefined`

apps/backend/src/lambdas/report.ts
```diff
@@ -1,16 +1,268 @@
...
+) => {
+  const cashDepositService = app.get(CashDepositService);
+  const posDepositService = app.get(PosDepositService);
+  const maxDate = parse(event.period.to, 'yyyy-MM-dd', new Date());
+  /* extract the month from the "to-date"*/
+  const minDate = new Date(getYear(maxDate), getMonth(maxDate), 1);
+
+  const dateRange = {
+    minDate: format(minDate, 'yyyy-MM-dd'),
+    maxDate: format(maxDate, 'yyyy-MM-dd'),
```
fw-noel, line 247: Please ensure that this isn't going back a day because of UTC

apps/backend/src/lambdas/report.ts
```diff
@@ -1,16 +1,268 @@
...
+    );
+
+  const currentCashPayments: {
+    payments: PaymentEntity[];
+    dateRange: DateRange;
+  }[] = [];
+  if (currentCashDeposits.length > 0) {
+    for (const l of locations) {
+      const cashDepositDatesForLocations =
+        await cashDepositService.findCashDepositDatesByLocation(
```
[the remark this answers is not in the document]
chelsea-EYDS: This is the only case where there is a request per location. The reason for this is that the deposit date range (the window of time between deposits that we query for the aggregated payments to match many to one to a deposit) is different for every locations. We could probably build a list of the most recent cash deposit dates and filter that by location, and then filter all of the matched payments based on the deposit date range, but it seemed overly complicated and the time for generating reports is fairly minimal. I am open to discussing though :)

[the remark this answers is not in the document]
chelsea-EYDS: Sorry, I guess here and also when retrieving the matched payments...

apps/backend/src/lambdas/report.ts
chelsea-EYDS: logged: 

`{ minDate: '2023-05-01', maxDate: '2023-05-25' } CAS DATE RANGE`

when generating a report for May 25th

apps/backend/src/lambdas/report.ts
chelsea-EYDS: I think since `pos_deposit_match` is an optional field on the PaymentEntity class it will always be possibly undefined? I tried to filter but had the same error. Happy to chat to get some direction on this...

apps/backend/src/lambdas/report.ts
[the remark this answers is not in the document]
chelsea-EYDS: resolved?

fw-noel (approved):

PR #126, opened 2023-06-07
[CCFPCM-0526] bugfix - heuristic 3

chelsea-EYDS:
[CCFPCM-0526](https://bcdevex.atlassian.net/browse/CCFPCM-0526)

Objective: 

- only match round three POS if deposit date is before payment date

### DIFF
--- a/apps/backend/src/reconciliation/pos-reconciliation.service.ts
+++ b/apps/backend/src/reconciliation/pos-reconciliation.service.ts
@@ -7,10 +7,7 @@ import {
   ReconciliationType,
 } from './types';
 import { MatchStatus } from '../common/const';
-import {
-  addBusinessDaysNoTimezone,
-  subtractBusinessDaysNoTimezone,
-} from '../common/utils/format';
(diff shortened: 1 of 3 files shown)


apps/backend/src/reconciliation/pos-reconciliation.service.ts
```diff
@@ -246,9 +243,6 @@ export class PosReconciliationService {
           // For round three, a deposit can be one business day apart from the payment
           // Could be the day before, or the day after, so we search those days for the amount
           if (posHeuristicRound === PosHeuristicRound.THREE) {
-            const dayAfterDateToFind = addBusinessDaysNoTimezone(dateToFind, 1);
-            deposits =
-              depositsWithAmount[`${dayAfterDateToFind}-${paymentMethod}`];
             if (!deposits?.length) {
```
fw-noel, line 246: I think this if case can be removed? Already checked in line 242

fw-noel (approved):

fw-noel (approved):

PR #127, opened 2023-06-07
[CCFPCM-0524] floating point decimal type precision loss bug

chelsea-EYDS:
[CCFPCM-0524](https://bcdevex.atlassian.net/browse/CCFPCM-0524)

Objective: 

- when aggregating payment and deposit sums we are losing .001 of a decimal in some cases, which is causing a false exception when matching for pos round four heuristics as well as cash payments to deposits
- implement decimal.js library to use the decimal type when summing the payment/deposit totals

### DIFF
--- a/apps/backend/src/reconciliation/cash-reconciliation.service.ts
+++ b/apps/backend/src/reconciliation/cash-reconciliation.service.ts
@@ -60,10 +60,8 @@ export class CashReconciliationService {
     deposit: CashDepositEntity
   ) {
     return (
-      Math.abs(
-        new Decimal(deposit.deposit_amt_cdn).toNumber() -
-          new Decimal(payment.amount).toNumber()
-      ) < 0.001
+      new Decimal(deposit.deposit_amt_cdn).toNumber() ===
+      new Decimal(payment.amount).toNumber()
     );
   }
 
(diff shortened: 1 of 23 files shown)


apps/backend/src/reconciliation/cash-reconciliation.service.ts
```diff
@@ -48,60 +49,65 @@ export class CashReconciliationService {
...
+  public checkAggregatedCashPaymentAmountToCashDepositAmount(
+    payment: AggregatedCashPayment,
     deposit: CashDepositEntity
   ) {
-    return Math.abs(deposit.deposit_amt_cdn - payment.amount) < 0.001;
+    return (
+      Math.abs(
+        new Decimal(deposit.deposit_amt_cdn).toNumber() -
+          new Decimal(payment.amount).toNumber()
+      ) < 0.001
```
fw-noel, line 66: Is < 0.001 still necessary with these changes?

apps/backend/src/reporting/cas-report/cas-report.ts
```diff
@@ -29,7 +30,8 @@ export class CasReport extends CasLocationReport {
     location: NormalizedLocation;
   }) {
     super(data.location);
+    this.settlement_date = data.settlement_date;
     this.card_vendor = data?.payment_method as string;
-    Object.assign(this, data);
```
fw-noel, line 33: How come this was removed?

fwkendall:
Reviewing at a high level - of the three standard js libraries (decimal, big, bignumber) this is the one that best suits our needs. A common pattern for handling simple cash amounts is to scale and store as an integer (e.g. 1.05$ would be stored as 105), but that won't suit our needs for division and for supporting foreign currency amounts. Looks good as a choice. 👍

fwkendall (approved):

fw-noel (approved):

PR #133, opened 2023-06-13
[CCFPCM-495] report updates 

chelsea-EYDS:
[CCFPCM-495](https://bcdevex.atlassian.net/browse/CCFPCM-495)

Objective: 
**POC**

- removed confusing deposit_date_range from the report
- renamed date to transaction_date (for pos payments/transactions)
- added deposit_date (for cash deposits plus matched cash payments)

- showing matched payments if the payment fiscal close date === report date (the only use case here would be for historical reports) 
- showing all relevant matched cash payments if there is a matched deposit where the cash deposit_date === report date

- shows exceptions for POS (looks back to the prev day for exceptions) 
- exceptions for cash (shows cash payments that have been thrown as an exception in which the fiscal_close_date is between the most recent deposit date for the location, and the two prev deposit dates prior...this might need to be tightened up a bit?)
^^ just a temporary solution until we implement the state change logging as a means to show relevant exceptions

- in order to remedy some of the confusion with the summary page I'm only showing the daily totals for match and exceptions, which should be sufficient since we're focused on the changed state for the report vs new unreconciled data

[report_june_7.xlsx](https://github.com/bcgov/PaymentCommonComponent/files/11741953/report_june_7.xlsx)


fw-noel (approved):
Looks ok!

PR #150, opened 2023-06-20
[CCFPCM-544] fix heap memory error for reports

chelsea-EYDS:
[CCFPCM-544](https://bcdevex.atlassian.net/browse/CCFPCM-544)

Objective: 

- When generating reports larger than a few days or weeks we are getting a JavaScript Heap out of memory error
- To resolve we are using a workbook stream so that the entire workbook does not need to be saved in memory prior to writing to a file
*note: deployed to test and verified upload (have since reverted test to main branch since this is not yet through PR)

### DIFF
--- a/apps/backend/src/excelexport/excelexport.service.ts
+++ b/apps/backend/src/excelexport/excelexport.service.ts
@@ -3,35 +3,26 @@
 import { Inject, Injectable, Logger } from '@nestjs/common';
 import { format, parse } from 'date-fns';
 import * as Excel from 'exceljs';
-import * as path from 'path';
 import { Stream } from 'stream';
+
 import { AppLogger } from '../logger/logger.service';
 import { Placement } from '../reporting/interfaces';
 import { S3ManagerService } from '../s3-manager/s3-manager.service';
 
 @Injectable()
 export class ExcelExportService {
-  private workbook: Excel.Workbook;
-
+  private workbook: Excel.stream.xlsx.WorkbookWriter;
+  private stream: Stream;
   constructor(
     @Inject(Logger) private readonly appLogger: AppLogger,
     @Inject(S3ManagerService) private readonly s3Manager: S3ManagerService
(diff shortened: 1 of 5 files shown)


apps/backend/src/reporting/reporting.service.ts
```diff
@@ -128,25 +125,16 @@ export class ReportingService {
       cashDeposits
     );
 
-    const startIndex = 2;
-
     this.excelWorkbook.addSheet(Report.DAILY_SUMMARY);
-
     this.excelWorkbook.addColumns(Report.DAILY_SUMMARY, dailySummaryColumns);
-
-    this.excelWorkbook.addRows(Report.DAILY_SUMMARY, summaryData, startIndex);
```
fw-noel, line 137: Intentional to remove addRows here? Not sure I understand

apps/backend/src/excelexport/excelexport.service.ts
```diff
@@ -40,35 +31,23 @@ export class ExcelExportService {
...
-              Key: `${filename}_${date}.xlsx`,
-              Bucket: `pcc-recon-reports-${process.env.RUNTIME_ENV}`,
-              Body: stream,
-              ContentType:
-                'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
-            })
-            .promise();
+      await this.s3Manager.s3
+        .upload({
+          Key: `${filename}_${date}.xlsx`,
```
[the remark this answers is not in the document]
chelsea-EYDS: Good call, I've updated this in the param that is passed in (to allow for single date or date range)

apps/backend/src/reporting/reporting.service.ts
chelsea-EYDS: Yes, intentional. Row data is last as they cannot be altered (no filters/style changes) after they are committed.

PR #152, opened 2023-06-21
[CCFPCM-522] Consistency updates for workflows and Node version, updated file sync naming

fwkendall:
[CCFPCM-522](https://bcdevex.atlassian.net/browse/CCFPCM-522)

Objective: 
- Applied consistency updates across the workflows
- Updated actions versions where applicable
- Locked Node at v18 in package.json to align with test workflow and terraform env (to be upgraded later since it's in LTS as of Feb, just aiming for predictability right now)
- Removed the param to prevent ZAP issue writing (docs were wrong? It expects different input however I can't find a counterexample to how I attempted to use it, which is following the docs, and I don't feel strongly about issue writing so I didn't invest more time investigating)
- Renamed the sync job to be clear it ONLY syncs files between S3 buckets, it does not parse or sync data

### DIFF
--- a/.github/workflows/ci-lint.yaml
+++ b/.github/workflows/ci-lint.yaml
@@ -1,4 +1,4 @@
-name: PR - Linting
+name: Lint backend
 
 on:
   workflow_dispatch:
@@ -7,7 +7,7 @@ on:
       - apps/backend/**
 
 jobs:
-  pr-checks:
+  lint:
     runs-on: ubuntu-latest
     timeout-minutes: 5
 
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-lint.yaml
```diff
@@ -1,4 +1,4 @@
-name: PR - Linting
+name: Lint backend
```
fwkendall, line 1: The renaming you see in this file (of the workflow and of the job below) will hopefully make things more readable for the PR checks

chelsea-EYDS (approved):
Looks good!

PR #153, opened 2023-06-21
[CCFPCM-525] check validity of footer

fw-noel:

Objective: 
We check TDI17 and TDI34 to ensure that the number of records and the total amount properly equal each other


apps/backend/src/parse/parse.controller.ts
```diff
@@ -163,19 +163,19 @@ export class ParseController {
           program,
           file.buffer
         ); // validating step
-        const fileToSave = await this.parseService.saveFileUploaded({
-          sourceFileType: fileType,
-          sourceFileName: fileName,
-          sourceFileLength: cashDeposits.length,
-          dailyUpload: daily,
-        });
+        // const fileToSave = await this.parseService.saveFileUploaded({
```
chelsea-EYDS, line 166: Not sure why we're commenting this out?

chelsea-EYDS (approved):
Looks good!

PR #154, opened 2023-06-22
[CCFPCM-525] reconcile only if there are valid files uploaded today

fw-noel:

Objective: 
As part of AC3 (though I might try to move that AC elsewhere), we want to prevent the reconciler from running if files haven't been uploaded today. Unfortunately, this is an issue if we're running the reconciler for different scenarios. For example, we haven't uploaded the files today yet, but we're testing the reconciler for the past.

One interesting thing. We can only run the reconciler with a max date of "yesterday". However, uploading files adds them in for "today", for data "yesterday". This will be addressed in future tickets, potentially [491](https://bcdevex.atlassian.net/browse/CCFPCM-491)

### DIFF
--- a/apps/backend/src/lambdas/reconcile.ts
+++ b/apps/backend/src/lambdas/reconcile.ts
@@ -1,6 +1,6 @@
 import { Logger } from '@nestjs/common';
 import { NestFactory } from '@nestjs/core';
-import { format, parse, subBusinessDays } from 'date-fns';
+import { format, parse, subBusinessDays, differenceInDays } from 'date-fns';
 import { AppModule } from '../app.module';
 import { MatchStatus } from '../common/const';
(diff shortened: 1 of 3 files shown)


apps/backend/src/lambdas/reconcile.ts
```diff
@@ -27,7 +28,22 @@ export const handler = async (event: ReconciliationConfigInput) => {
   const posDepositService = app.get(PosDepositService);
   const locationService = app.get(LocationService);
   const reportingService = app.get(ReportingService);
+  const parseService = app.get(ParseService);
   const appLogger = app.get(Logger);
+
+  console.log(event.bypass_parse_validity);
```
chelsea-EYDS, line 34: console.log

chelsea-EYDS (approved):

PR #155, opened 2023-06-23
[FIX] error message for preventing reconciler

fw-noel:

Objective: 
Just an error message change

### DIFF
--- a/apps/backend/src/lambdas/reconcile.ts
+++ b/apps/backend/src/lambdas/reconcile.ts
@@ -39,7 +39,9 @@ export const handler = async (event: ReconciliationConfigInput) => {
     }
     const daily = await parseService.getDailyForRule(rule, new Date());
     if (!daily?.success) {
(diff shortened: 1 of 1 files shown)


apps/backend/src/lambdas/reconcile.ts
```diff
@@ -39,7 +39,9 @@ export const handler = async (event: ReconciliationConfigInput) => {
     }
     const daily = await parseService.getDailyForRule(rule, new Date());
     if (!daily?.success) {
-      throw new Error('There are still invalid files for thie date');
+      throw new Error(
+        `Incomplete dataset for this date ${new Date()}. Please check the uploaded files.`
```
fwkendall, line 43: my only nitpick is it would be nice to store the date as a variable instead of producing it twice (super edge case it's midnight) - no worries if we get this in the next PR :)

fwkendall (approved):

PR #157, opened 2023-06-27
[CCFPCM-519] Clear DB data and S3 jobs

fwkendall:
[CCFPCM-519](https://bcdevex.atlassian.net/browse/CCFPCM-519)

Objective: 
- Added two jobs, one that resets status and one that clears all key data (DB and S3)
- The new Makefile command for the S3 buckets is targeting subfolders
- I'll need to run these to confirm they work in GHA env (I believe the necessary connection vars are already available to the make commands - but need to confirm) - got the all-clear from Qudsia already to wipe dev DB as needed for testing
- We also have the option to configure approvers for these workflows in GH if we want to

### DIFF
--- a/.github/workflows/clear-all-data-dev.yaml
+++ b/.github/workflows/clear-all-data-dev.yaml
@@ -1,0 +1,49 @@
+name: Clear all data - DEV
+on:
+  workflow_dispatch:
+env:
+  ENV_NAME: dev
+
+jobs:
+  clear-all-data:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    environment:
+      name: dev
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Cache yarn
+        uses: actions/cache@v3
+        with:
+          path: ./.yarn/cache
+          key: pcc-backend
+          restore-keys: pcc-backend
+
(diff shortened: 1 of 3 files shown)


.github/workflows/clear-all-data-dev.yaml
```diff
@@ -0,0 +1,49 @@
...
+        run: cp ./.config/.env.example ./.env
+
+      - name: Clear S3 bucket
+        run: make aws-empty-s3-bucket-dev
+
+      - name: Open DB tunnel
+        run: make open-db-tunnel
+
+      - name: Clear DB
+        run: make clear-database
```
chelsea-EYDS, line 49: This looks good. I think we might be missing a couple of things from the make clear-database (new items that should be cleared...)

`delete from program_daily_upload`

and

`delete from file_uploaded`

I am happy to create a separate PR for these (I should have added them in previously, it is not part of this for this PR but just FYI)

fwkendall: Oh shoot, I should have reviewed that more closely. I can add in another commit!

fwkendall: Added

chelsea-EYDS (approved):

PR #159, opened 2023-06-27
[CHORE] change migrations class name to uppercase

chelsea-EYDS:

Objective: 
- change this to remove the warning from sonar scan


fw-noel:
I would actually just ignore migrations in sonar

PR #168, opened 2023-07-10
[FIX] added missing uploaded on date

chelsea-EYDS:

Objective: 

- added the uploaded on field to the transaction table
- include in the report


fw-noel (commented):
Aside from my comment, I think I'd just like to ask you to rename the column to `created_date` even if the report is `uploaded_date`. It's a) more common and b) accounts for API uploads

fw-noel (approved):

PR #171, opened 2023-07-14
[CHORE] updated readme

chelsea-EYDS:

Objective: 
- updated readme (after getting project running on my new laptop :D )


fw-noel (commented):
Not sure I agree with consolidating in one place or removing the reconciliation piece

chelsea-EYDS:
> Not sure I agree with consolidating in one place or removing the reconciliation piece

Reason for consolidating: easier to read and follow when viewing in the repo and trying to follow along during project set up. 

Reason for removing the reconciliation piece: accuracy. (It needs to be re-written but I thought nothing was better than something that was inaccurate/out of date)

PR #172, opened 2023-07-14
[CCFPCM-577] Require IMDSv2 for bastion host config

fwkendall:
[CCFPCM-577](https://bcdevex.atlassian.net/browse/CCFPCM-577)

Objective: 
- Addresses the high sev alert in Security Hub
- Must specify this metadata to limit IMDS to only support v2, not also v1 (default setting allows both)
- Note I haven't applied this change yet :) I decided it would be best to keep scope small for this fix, but open to hearing we should address other things as part of this same PR


fw-noel (approved):
I'm up for fixing the high / crit ones in increments and adding tech debt for the rest

PR #173, opened 2023-07-18
[CCFPCM-491] required files, alert destinations

fw-noel:
[CCFPCM-491](https://bcdevex.atlassian.net/browse/CCFPCM-491)

Objective: 
This addressed AC#1, AC#2 for 491. Using the database, we can configure:
- Filenames in a more fine-grained way, so that if, for example, a program requires two different TDI17s, we can account for that
- Alerts in a more fine-grained way, so that we can configure who gets what email

### DIFF
--- a/apps/backend/src/notification/entities/alert-destination.entity.ts
+++ b/apps/backend/src/notification/entities/alert-destination.entity.ts
@@ -1,0 +1,41 @@
+import {
+  Column,
+  CreateDateColumn,
+  PrimaryGeneratedColumn,
+  Entity,
+  ManyToOne,
+  JoinColumn,
+} from 'typeorm';
+import { FileIngestionRulesEntity } from '../../parse/entities/file-ingestion-rules.entity';
+import { ProgramRequiredFileEntity } from '../../parse/entities/program-required-file.entity';
+
+@Entity('alert_destination')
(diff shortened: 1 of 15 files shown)


apps/backend/src/notification/entities/alert-destination.entity.ts
```diff
@@ -0,0 +1,41 @@
...
+export class AlertDestinationEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @CreateDateColumn({ name: 'created_date' })
+  createdDate: Date;
+
+  /**
+   * Only one of the next three must be truthy
+   */
```
chelsea-EYDS, line 22: should 'allAlerts' and/or 'requiredFile' be optional? (ie: 'rule?')

chelsea-EYDS (approved):
Looks Good!

apps/backend/src/notification/entities/alert-destination.entity.ts
fw-noel: Technically we don't have validation to ensure that only one of the three is truthy, but yeah they should all be optional. Happy to get on a call to discuss

chelsea-EYDS (approved):

PR #177, opened 2023-07-23
[CHORE] clean up unused files

chelsea-EYDS:

Objective: 

- clean up unused files to reduce build zip size


fw-noel (approved):
Before merging, I'd ensure that the ones that might be referred to in the future are in the Sharepoint folder in a good spot, and maybe add to the main readme if there's any ones that might be pertinent, like the

PR #182, opened 2023-07-24
[CCFPCM-0575] reconciliation to trigger report

chelsea-EYDS:
[CCFPCM-0575](https://bcdevex.atlassian.net/browse/CCFPCM-0575)

Objective: 
- Reconciliation Lambda to trigger report output
- TODO: set the daily file check lambda to trigger the reconciliation once daily per LoB (if all files present, otherwise, alert)
- TODO: move this outside of the lambda function and use the destination to trigger the lambda instead
- **note:** the reconciled on, in progress on, and uploaded on dates are still a bit messed up. We will need to use only the realtime dates once automation is in place, but if we run a batch reconciliation for a specific date and then try to generate the report it will break, as the report will include all items reconciled on that date (so if you reconcile the entire dataset in one go you will be generating a report for the full dataset)

### DIFF
--- a/apps/backend/src/lambdas/reconcile.ts
+++ b/apps/backend/src/lambdas/reconcile.ts
@@ -186,5 +186,5 @@ export const handler = async (event: ReconciliationConfigInput) => {
   console.table(cashReport);
 
   //TODO use this only for local and set the reconciliation lambda destination to trigger the report
-  await reportHandler(event);
+  isLocal && (await reportHandler(event));
 };
--- a/apps/backend/src/lambdas/reconcile.ts
+++ b/apps/backend/src/lambdas/reconcile.ts
@@ -1,7 +1,8 @@
 import { Logger } from '@nestjs/common';
 import { NestFactory } from '@nestjs/core';
-import { format, parse, subBusinessDays } from 'date-fns';
+import { differenceInDays, format, parse, subBusinessDays } from 'date-fns';
 import { FindOptionsOrderValue } from 'typeorm';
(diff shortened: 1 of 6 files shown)


apps/backend/src/lambdas/reconcile.ts
```diff
@@ -125,42 +132,44 @@ export const handler = async (event: ReconciliationConfigInput) => {
...
-        filtered[filtered.indexOf(date) - 2] ??
-        format(reconciliationDates.start, 'yyyy-MM-dd');
+        filtered[filtered.indexOf(date) - 2] ?? format(startDate, 'yyyy-MM-dd');
 
       const dateRange = {
         minDate: previousCashDepositDate,
         maxDate: currentCashDepositDate,
       };
 
+      //TODO replace current cashDepositDate with actual date
```
fw-noel, line 158: Ticket for this todo?

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -175,4 +184,7 @@ export const handler = async (event: ReconciliationConfigInput) => {
   const cashReport = await reportingService.reportCashMatchSummaryByDate();
   appLogger.log('\n\n=========Cash Summary Report: =========\n');
   console.table(cashReport);
+
+  //TODO use this only for local and set the reconciliation lambda destination to trigger the report
```
fw-noel, line 188: We probably shouldn't merge this in until this TODO is done?

fw-noel (approved):

PR #183, opened 2023-07-24
[CCFPCM-0572][CCFPCM-0340] s3trigger to parser / alerts lambda (WIP)

chelsea-EYDS:
[CCFPCM-0340](https://bcdevex.atlassian.net/browse/CCFPCM-0340)

Objective:


apps/backend/fixtures/lambda/parser.json
```diff
@@ -1 +1 @@
-{ "eventType": "all" }
\ No newline at end of file
+{ "Records": [] }
```
fw-noel, line 1: ?

apps/backend/src/lambdas/dailyfilecheck.ts
```diff
@@ -0,0 +1,71 @@
...
+        await mailService.sendEmailAlertBulk(
+          MAIL_TEMPLATE_ENUM.FILES_MISSING_ALERT,
+          alertDestinations.map((ad) => ({
+            toEmail: ad,
+            message: errors.join(' '),
+          }))
+        );
+      }
+    }
+    //TODO trigger reconciliation if has all files
```
fw-noel, line 65: I disagree that this should trigger reconciliation. I think it needs to be wholly separate.

Besides, what scenario would there be where the parser wouldn't trigger the reconciliation but this would?

apps/backend/src/parse/parse.controller.ts
```diff
@@ -111,135 +110,14 @@ export class ParseController {
...
-      const message =
-        err instanceof Error
-          ? err.message
-          : `Error with processing ${fileName}`;
-      this.appLogger.log(message);
-      throw new BadRequestException({ message });
-    }
+    //TODO call parse service to process file event
+    return {
+      fileName,
```
fw-noel, line 115: This may be more appropriate to return?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

apps/backend/src/parse/parse.service.ts
```diff
@@ -248,7 +239,7 @@ export class ParseService {
    * @returns List of uploaded files
    */
   async getAllFiles(): Promise<FileUploadedEntity[]> {
-    return this.uploadedRepo.find();
+    return await this.uploadedRepo.find();
```
fw-noel, line 242: Unnecessary, I'm pretty sure. Since `repo.find` is an async function, the await goes to the top.

Same with the ones below (saveFileUploaded)

apps/backend/fixtures/lambda/parser.json
chelsea-EYDS: This is the shape of the data that we get when the parser lambda is triggered by updates in the S3 bucket. If we receive an empty Records object then we will do as we did before and check the S3 bucket for any discrepancies between the bucket and the db and then parse. (This is so we can still call 'make parse' locally...)

apps/backend/src/lambdas/dailyfilecheck.ts
chelsea-EYDS: This is the part of the automation flow that I think we still need to discuss a bit more as a team. Are we scheduling reconciliation, and then using that trigger to first run this daily check, or, are we scheduling the daily check, and if successful, running reconciliation?

apps/backend/src/parse/parse.service.ts
```diff
@@ -263,84 +254,254 @@ export class ParseService {
...
-    return this.programDailyRepo.findOne({
-      relations: ['rule', 'files'],
-      where: {
-        dailyDate: format(date, 'yyyy-MM-dd'),
-        rule: {
-          id: rule.id,
-        },
-      },
-    });
+  async parseFile(fileKey: string) {
```
fw-noel, line 307:
```suggestion
  async parseFileFromS3(fileKey: string) {
```

fw-noel (approved):

fw-noel (approved):

PR #186, opened 2023-07-26
[CCFPCM-565] fix pos amounts for reports page 3

fw-noel:
[CCFPCM-565](https://bcdevex.atlassian.net/browse/CCFPCM-565)

Objective: 
The left join on the payment method was adding more rows than necessary because it was left-joining the 1:X relationships on the entity. This PR makes it closer to Karim's query


chelsea-EYDS (approved):
nice

PR #187, opened 2023-07-27
[CCFPCM-0340]  parser -> reconcile

chelsea-EYDS:
[CCFPCM-0340](https://bcdevex.atlassian.net/browse/CCFPCM-0340)

Objective: 

- automated reconciliation: attempt to run recon whenever a file is parsed. Do not run if all daily files are not present
- update to parser to use file metadata to set the daily date and save in the db (also saving the actual date of upload in order to easily compares these dates in case of a late file etc)
- allow for configurable manual override for reconciliation
- set up default configuration for reconciliation for the daily job 

Testing Notes:

- have deployed to dev
- parsing will trigger reconciliation for the current date as max date and the current date minus 31 for the min date
- check the logs to ensure that parsing triggers reconciliation
- if you want to over ride the reconciliation inputs you can do that in the test section of the lambda
- after reconciliation has completed, a report will be uploaded in the S3 bucket. 
- to test e2e you can delete all or some files from the S3 bucket, then manually trigger the sync job for prod S3->devS3 
- Once parsing has finished, check the reconciler logs to ensure reconciliation has started with the default params
- Once the reconciler is finished check the reports logs 
- After the report lambda has completed it will output the report to the S3 bucket


apps/backend/src/database/migrations/1690503718109-daily-date.ts
```diff
@@ -0,0 +1,42 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migration1690503718109 implements MigrationInterface {
+  name = 'daily-date-migration1690503718109';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "pos_deposit" ADD "file_created_date" TIMESTAMP`
```
fw-noel, line 8: Need an explanation for this migration.
- If each deposit / transaction is associated with a file, why the repeated column on each?
- Is file_created_date meaningfully different from daily_date?

apps/backend/src/lambdas/helpers.ts
```diff
@@ -0,0 +1,60 @@
+import { parse, subBusinessDays } from 'date-fns';
+import { Ministries } from '../constants';
+
+export const setReconciliationDates = (
```
fw-noel, line 4: Comment block explanation

apps/backend/src/lambdas/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+  event: unknown,
+  maxNumDaysToReconcile: number
```
fw-noel, line 6: This seems like it's more `numDaysToReconcile` than a `max` of some sort, am I misunderstanding?

apps/backend/src/lambdas/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+    currentDate: new Date(),
+    ministry: Ministries.SBC,
+    byPassFileValidity: false,
+  };
+};
+
+/**
+ * Extract the file date from the file name
+ */
+export const extractDateFromTXNFileName = (fileName: string): Date => {
```
fw-noel, line 44: My concern about these functions is that they expect filenames that conform exactly to what we want. The two ways to mitigate that:
- Add an example of the filename in the comment block
- A trycatch / error if it does not conform to what we expect. Same thing if the date is not parseable

apps/backend/src/database/migrations/1690503718109-daily-date.ts
chelsea-EYDS: I just thought that it should be included in the file metadata since we're extracting that from the file itself...

apps/backend/src/lambdas/helpers.ts
chelsea-EYDS: I've removed the date from filename stuff for bcm. For garms, we don't really have a choice, but I have included a validation to reject the file in the case of an incorrect filename.

apps/backend/src/database/migrations/1690503718109-daily-date.ts
chelsea-EYDS: (I think there was talk of removing the metadata column altogether? I'm just using this for the reporting uploaded date but can remove and refactor to use the daily date instead)

apps/backend/src/lambdas/helpers.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Extract the file date from the file name
+ * exmaple filename: 'sbc/SBC_SALES_2026_03_03_23_17_37.JSON'
+ */
+export const extractDateFromTXNFileName = (fileName: string): Date => {
+  try {
+    const name = fileName.split('/')[1].split('.')[0];
+    const date = name.replace('SBC_SALES_', '').replace(/[_]/gi, '-');
+    return new Date(date.slice(0, 10));
+  } catch (err) {
+    throw new Error(`Error extracting date from file name: ${fileName}`);
```
fw-noel, line 63: I don't believe new Date('anyinvaliddatestring'); throws an error so I don't think this can actually be caught. LMK if I'm wrong.

I'll suggest using date-fns parse instead? I **think** that throws a RangeError for an incorrectly formatted date.

apps/backend/src/lambdas/helpers.ts
```diff
@@ -0,0 +1,80 @@
...
+};
+/**
+ * Validates the filename from SBC Garms. Example filename: 'sbc/SBC_SALES_2026_03_03_23_17_37.JSON',
+ * @param filename
+ * @returns
+ *
+ */
+export const validateSbcGarmsFileName = (filename: string): boolean => {
+  const reg =
+    /sbc\/\SBC_SALES_(20)\d{2}_(0[1-9]|1[0,1,2])_(0[1-9]|[12][0-9]|3[01])_\d{2}_\d{2}_\d{2}.JSON/gi;
```
fw-noel, line 74: Throw this in a constants file please!

apps/backend/src/database/migrations/1690503718109-daily-date.ts
```diff
@@ -0,0 +1,47 @@
...
+      `ALTER TABLE "cash_deposit" ALTER COLUMN "file_created_date" TYPE date USING "file_created_date"::date`
+    );
+    await queryRunner.query(
+      `ALTER TABLE "transaction" ALTER COLUMN "file_created_date" TYPE date USING "file_created_date"::date`
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "file_uploaded" DROP COLUMN "file_created_date"`
```
fw-noel, line 29: This should be transaction

apps/backend/src/database/migrations/1690503718109-daily-date.ts
```diff
@@ -0,0 +1,47 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migration1690503718109 implements MigrationInterface {
```
fw-noel, line 3: Overall suggestion: change all these to `parsed_date` for each table, do not do it for `file_uploaded` entity

fw-noel (approved):

PR #189, opened 2023-07-31
[CCFPCM-610] batch-reconciliation

chelsea-EYDS:
[CCFPCM-610](https://bcdevex.atlassian.net/browse/CCFPCM-610)

**Objective:** 

In order to run the reconciliation on the entire dataset as though it had run daily we should use the file metadata and filename (for SBC json files) to extract the date and run reconciliation based on that date. 

**How To Run On local:**

- Parsing, reconciliation and reports are now all automated by default. When you run sync/parse the entire dataset will be reconciled (job is >1 hr but should only need to run once per env unless you drop your db)

    - to test: ‘make drop', ‘make migrations-run’, ‘make sync’, 'make parse’ 

- Reconciliation can be run manually (only runs for the specified dates, does not batch):

    - 'make reconcile' and update reconcile.json

- Automated reconciliation can be turned off:

    - set DISABLE_AUTOMATED_RECONCILIATION=true

- Automated reports can be set to false in the reconciliation input:

   - "generateReport": false 

- Reports can be generated manually 

   - run 'make report' after updating report.json

- If you do not want to re-parse your data, or you only want to batch reconcile some of the data:

   - run 'make batch-reconcile' (uses default inputs for  Jan 1 to current date, these can be overridden in batch.json)

**How It Runs On Dev/Test:**

- Whenever data is dropped into the S3 bucket, either manually or via the sync job, it will be parsed and reconciled and will generate a report

- On dev, if you want to use the manual overrides you should drop your test data into the db and allow it to be parsed. Then, run the queries to reset the status to the pre-reconciled state. Now you can use the test tab in the lambdas to run manually.

### DIFF
--- a/apps/backend/src/parse/parse.service.ts
+++ b/apps/backend/src/parse/parse.service.ts
@@ -8,8 +8,9 @@ import {
 } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { S3Event, S3EventRecord } from 'aws-lambda';
+import { SNS } from 'aws-sdk';
 import { validateOrReject, ValidationError } from 'class-validator';
-import { format } from 'date-fns';
+import { format, parse, subBusinessDays } from 'date-fns';
 import { Repository } from 'typeorm';
 import _ from 'underscore';
 import { CashDepositDTO, CashDepositsListDTO } from './dto/cash-deposit.dto';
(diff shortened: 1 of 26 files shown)


apps/backend/src/parse/parse.service.ts
```diff
@@ -325,16 +305,65 @@ export class ParseService {
       // Parse & Save only files that have not been parsed before
       for (const filename of finalParseList) {
         this.appLogger.log(`Parsing ${filename}..`);
+        const isLocal = process.env.RUNTIME_ENV === 'local';
```
fw-noel, line 308: I'd pull this variable way up, like the beginning of the function or even class

fw-noel (commented):
Looks ok! I would maybe throw into the readme why there's lambdas for reconciliation and why

fw-noel (approved):

PR #196, opened 2023-08-08
[CCFPCM-441] base auth guard

fw-noel:
[CCFPCM-441](https://bcdevex.atlassian.net/browse/CCFPCM-441)

Objective: 
Global auth guard applied to all requests but for healthcheck


apps/backend/src/common/guards/auth.guard.ts
```diff
@@ -0,0 +1,124 @@
...
+  Injectable,
+  UnauthorizedException,
+} from '@nestjs/common';
+import { Reflector } from '@nestjs/core';
+import axios, { AxiosInstance } from 'axios';
+import { Request } from 'express';
+import jwt from 'jsonwebtoken';
+import { PUBLIC_ENDPOINT } from './../decorators/publicEndpoint.decorator';
+
+let axiosInstance: AxiosInstance;
```
chelsea-EYDS, line 14: Curious why we are defining axios instance outside of the class but assigning the value in the constructor?

chelsea-EYDS (approved):
Looks good! One question re: the axios instance in the auth guard... should this be an instance variable? Or if a  public field on the class should the value be assigned outside of the constructor? (I don't think it should affect the functionality but just curious )

fw-noel:
Good catch!

chelsea-EYDS (approved):
Nice, looks great!

chelsea-EYDS (approved):

PR #197, opened 2023-08-08
[FIX] fix bucket name for local static data

chelsea-EYDS:

Objective: 

- fix bucket name for local master data


fw-noel (approved):
I'm not sure why but I see no issues with it

PR #198, opened 2023-08-08
[FIX] fix SNS topics to work in each env

chelsea-EYDS:

Objective: 

- removed hardcoded topic env vars 
- using account id from the proper env


fw-noel (commented):
Imo, it would be cleaner to define the hardcoded stuff in the variables

fw-noel (approved):

PR #201, opened 2023-08-10
[CCFPCM-584] SQ error fix + Logger update

chelsea-EYDS:
CCFPCM-584[CCFPCM-0000](https://bcdevex.atlassian.net/browse/CCFPCM-584)

Objective: 

- resolve SQ bug
- update logger

### DIFF
--- a/apps/backend/src/deposits/cash-deposit.service.ts
+++ b/apps/backend/src/deposits/cash-deposit.service.ts
@@ -1,4 +1,4 @@
-import { Inject, Injectable, Logger } from '@nestjs/common';
+import { Inject, Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { addBusinessDays, parse, subBusinessDays } from 'date-fns';
 import {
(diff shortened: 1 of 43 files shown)


apps/backend/src/deposits/cash-deposit.service.ts
```diff
@@ -18,11 +18,13 @@ import { AppLogger } from '../logger/logger.service';
...
 @Injectable()
 export class CashDepositService {
+  private readonly appLogger;
   constructor(
-    @Inject(Logger) private readonly appLogger: AppLogger,
     @InjectRepository(CashDepositEntity)
     private cashDepositRepo: Repository<CashDepositEntity>
-  ) {}
+  ) {
+    this.appLogger = new AppLogger(CashDepositService.name);
```
fw-noel, line 26: In this PR, we are instantiating appLogger for every class. Which I think is an okay solution, but I do prefer the injectable I think, https://docs.nestjs.com/techniques/logger#injecting-a-custom-logger

fw-noel (approved):

PR #202, opened 2023-08-11
[CCFPCM-0578] pkg updates

chelsea-EYDS:
[CCFPCM-0578](https://bcdevex.atlassian.net/browse/CCFPCM-0578)

Objective: 

- upgrade all pks to latest v
- upgrade to aws sdk v3 
- update code where needed 
- addressed faker deprecation warnings on test files

Screenshot 2023-08-14 at 10 12 46 AM


fw-noel (approved):

apps/backend/src/app.module.ts
```diff
@@ -34,23 +32,7 @@ import { TransactionModule } from './transaction/transaction.module';
...
-        process.env.RUNTIME_ENV === 'ci'
-          ? {
-              endpoint: process.env.AWS_ENDPOINT || 'http://localhost:9000',
-              region: 'ca-central-1',
-              s3ForcePathStyle: true,
-            }
-          : {}),
-      },
-      services: [S3, SNS],
+      ignoreEnvFile: process.env.NODE_ENV === 'production',
```
fwkendall, line 35: why do we no longer ignore the env file in dev/test?

apps/backend/src/lambdas/helpers.ts
```diff
@@ -28,25 +30,31 @@ export const validateSbcGarmsFileName = (filename: string): boolean => {
  * @param message
  * @returns
  */
-export const generateLocalSNSMessage = (message: unknown) => {
+/**
+ * Used to generate a local SNS message for running lambdas locally
```
fwkendall, line 34: I'm wondering if this wasn't working until now, given the defaults you had to add to sns-manager.service.ts ?

apps/backend/src/app.module.ts
chelsea-EYDS: NODE_ENV==='production' applies to dev/test. The RUNTIME_ENV differentiates dev/test/prod

chelsea-EYDS: Previously we were using the "nest-aws-sdk" (open source pkg) to set those defaults in the app module and thety were available in the services using the "@InjectAWS" decorator. That pkg was not compatible with v3. There is a similar pkg we could use instead: [here](https://www.npmjs.com/package/aws-sdk-v3-nest) but it didn't seem necessary....

apps/backend/src/lambdas/helpers.ts
chelsea-EYDS: Previously we were using the "nest-aws-sdk" (open source pkg) to set those defaults in the app module and they were available in the services using the "@InjectAWS" decorator. That pkg was not compatible with v3. There is a similar pkg we could use instead: [here](https://www.npmjs.com/package/aws-sdk-v3-nest) but it didn't seem necessary....

chelsea-EYDS: oh, also, this doesn't _really_ generate a local SNS message...I think we would have to use local stack for that. This will just call the handler with the same type as the SNS message that is passed in on dev/test/prod

apps/backend/src/app.module.ts
fwkendall: Perfect, I was hoping that was how we were using it but given how it was written before I wasn't entirely sure!

fwkendall (dismissed):

fwkendall (approved):

fwkendall (approved):

PR #203, opened 2023-08-15
[CCFPCM-0000] Bundle dependencies as lambda layer 

simensma-fresh:

Objective: 
**Reduce lambda bundle size**
Direct uploads of lambda deployment packages  are limited to 50mb. This PR adds support for bundling and node dependencies as a separate lambda layer that can be shared between the lambdas, and fixes this issue.

**Changes**
- Added s3 bucket for build artifacts. Lambda layers face the same 50mb limit for direct uploads, but a 250mb limit for layers deployed via an s3 bucket
- Updated build scripts to build two separate artifacts (<GIT_COMMIT_SHA>/backend/backend.zip and <GIT_COMMIT_SHA>/layer/nodejs.zip) that are uploaded to s3 
- Updated Github Actions workflows to upload artifacts and publishing a new lambda layer before deploying code changes
- Moved common deployment scripts into `deploy.sh` to keep things neat.

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+#####
+# Publishes the latest built layer to s3 and publishes a new lambda layer version
+# Usage: aws-publish-layer
+#####
+function aws-publish-layer() {
+	aws lambda publish-layer-version \
+		--layer-name "ppc-dependencies" \
+		--content S3Bucket=$APP_SRC_BUCKET,S3Key=$COMMIT_SHA/layer/nodejs.zip \
+		--compatible-runtimes nodejs18.x \
+		--region ca-central-1 \
+		--output text \
+		--query 'LayerVersionArn' \
+		--no-cli-pager
+}
+
+function aws-get-latest-layer-version() {
+    aws lambda list-layer-versions \
+            --layer-name ppc-dependencies \
+            --region ca-central-1 \
+            --query 'LayerVersions[0].LayerVersionArn'
+}
+
+#####
+# Updates the given lambda function to the latest code and layer version
+# Usage: aws-update-function-code <function-name>
+#####
+function aws-update-function-code() {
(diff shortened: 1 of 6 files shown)


chelsea-EYDS:
It looks good to me! I tried to fetch but I wasn't able to.... could we run `make tag-dev` just as a sanity check? Thank you so much for doing this!

chelsea-EYDS (approved):
Looks great!

bin/deploy.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+#####
+# Publishes the latest built layer to s3 and publishes a new lambda layer version
+# Usage: aws-publish-layer
+#####
+function aws-publish-layer() {
+  aws lambda publish-layer-version \
+    --layer-name "ppc-dependencies" \
```
chelsea-EYDS, line 9: _extremely_ minor, but could we call this 'pcc-dependencies'?

apps/backend/package.json
```diff
@@ -88,7 +88,9 @@
     "lint-staged": "13.1.0",
     "prettier": "2.8.3",
     "supertest": "6.3.3",
+    "terser-webpack-plugin": "^5.3.9",
```
chelsea-EYDS, line 91: one question: In previous projects, I have been asked to remove the caret prefix on the package version in order to lock it to the exact minor/patch version. Is this still a preferred practice? It seems like leaving the caret prefix and allowing for minor/patch updates would not be breaking?

chelsea-EYDS (approved):
awesome, thank you!!

apps/backend/package.json
simensma-fresh: Updated to lock it down.

Yeah I try to lock them down as well. *Most* of the time minor / patch updates don't break things, buuuut from experience not always 😅

chelsea-EYDS (approved):

PR #204, opened 2023-08-16
[FIX] NaN values in report

chelsea-EYDS:

Objective: 

- replace NaN with null
- use correct date

### DIFF
--- a/apps/backend/src/reconciliation/pos-reconciliation.service.ts
+++ b/apps/backend/src/reconciliation/pos-reconciliation.service.ts
@@ -1,10 +1,5 @@
 import { Injectable, Inject } from '@nestjs/common';
-import {
-  addBusinessDays,
-  differenceInBusinessDays,
-  differenceInMinutes,
-  parse,
-} from 'date-fns';
+import { differenceInBusinessDays, differenceInMinutes } from 'date-fns';
 import Decimal from 'decimal.js';
 import {
   PosDepositsAmountDictionary,
@@ -492,13 +487,7 @@ export class PosReconciliationService {
     const updatedPayments = inProgressPayments.map((itm) => ({
       ...itm,
       timestamp: itm.timestamp,
-      reconciled_on:
-        process.env.RUNTIME_ENV === 'production'
-          ? currentDate
-          : addBusinessDays(
-              parse(itm.transaction.transaction_date, 'yyyy-MM-dd', new Date()),
-              2
-            ),
+      reconciled_on: currentDate,
       status: MatchStatus.EXCEPTION,
     }));
     const paymentExceptions = await this.paymentService.updatePaymentStatus(
@@ -508,15 +497,10 @@ export class PosReconciliationService {
     const updatedDeposits = inProgressDeposits.map((itm) => ({
       ...itm,
       timestamp: itm.timestamp,
-      reconciled_on:
-        process.env.RUNTIME_ENV === 'production'
-          ? currentDate
-          : addBusinessDays(
-              parse(itm.transaction_date, 'yyyy-MM-dd', new Date()),
-              2
-            ),
+      reconciled_on: currentDate,
       status: MatchStatus.EXCEPTION,
     }));
+
     const depositExceptions = await this.posDepositService.updateDepositStatus(
       updatedDeposits // TO EXCEPTION
     );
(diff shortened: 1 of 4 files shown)


apps/backend/fixtures/lambda/batch.json
```diff
@@ -1,8 +1,8 @@
 {
   "period": {
-    "from": "",
+    "from": "2023-06-01",
     "to": ""
   },
   "program": "SBC",
-  "generateReport": true
+  "reportEnabled": false
```
fwkendall, line 7: I'm confused about the rename here?

fwkendall: I see it's also renamed in reconcile.json and in batch-reconcile.ts, was this just for naming clarity?

apps/backend/src/lambdas/batch-reconcile.ts
```diff
@@ -1,16 +1,21 @@
...
 import { AppLogger } from '../logger/logger.service';
 
 import { SnsManagerService } from '../sns-manager/sns-manager.service';
-
-/* eslint-disable @typescript-eslint/no-explicit-any */
-export const handler = async (event: HandlerEvent, _context?: Context) => {
+/**
+ * Enables batch reconciliation by iterating over the dates for a given period.
+ * Use batch.json to configure locally.
+ * Not used in production.
```
fwkendall, line 14: Possible tech debt to log - if we don't intend this to be used in prod I think we might want to add a check and failure state here for the prod env so it can't fire by accident

apps/backend/fixtures/lambda/reconcile.json
```diff
@@ -3,12 +3,9 @@
     {
       "Sns": {
         "Message": {
-          "period": {
-            "from": "2023-01-01",
-            "to": "2023-01-02"
-          },
-          "program": "SBC", 
-          "generateReport": true, 
+          "reconciliationMaxDate": "2023-08-14",
```
fwkendall, line 6: I'm confused why this PR is touching the scope of removing the start date from reconciliation (switching from period to only a max date).

fw-noel: We've had some discussions where we felt a start date wasn't super necessary anymore - IMO if we're making an attempt to reconcile, we should be reconciling the data we have regardless of when they started. The date range was really for supporting batch reconcile, which is something I think we're focusing less on, opting to focus on the real world scenario of saying "reconcile up until this date" (typically "yesterday").

fwkendall: We 100% have discussed changing this but it's not in the PR description or part of the ticket, so I'm not super sure on the state of things - did we already remove other code related to this and this PR is some missed cleanup? Or is this PR the functional change and we'll need corresponding documentation/signoff of that?

apps/backend/fixtures/lambda/batch.json
chelsea-EYDS: Yes. generateReport sounded more the action of actually generating it, so I have used this for the func name that is called if "reportEnabled" is true.

apps/backend/src/lambdas/batch-reconcile.ts
chelsea-EYDS: Hmm, good call. I mean, it _could_ be used in Prod, but it would have to be done so with intention. The only way to trigger this one (aside from a code-change) is to log into the prod aws lambda and "test" the lambda manually. I believe access to this env is limited (and whoever triggers it would also need to include correct inputs in order to run.) We may use this on tools/test/dev to test out new heuristics layers or code changes to reconciliation applied to the entire db, or a large chunk of it.

apps/backend/fixtures/lambda/reconcile.json
chelsea-EYDS: The requirements for running reconciliation in production only require one date, but for internal and testing purposes we still need the flexibility to reset data and reconcile in batches. I should have made this change with the initial work on splitting out the test/batch logic from the production flow but I missed the opportunity then. 

TBH we might want to remove this input altogether as reconciliation should really only be called from the parser directly in prod, and/or from the batch-reconcile handler if batching in dev/test. 

The only use case I see for keeping these inputs is to individually trigger the lambda for debugging purposes.

apps/backend/src/lambdas/batch-reconcile.ts
fwkendall: I guess it's possible we would need this available on prod to recover from multiple days of missing files... maybe let's leave this for now until we have more clarity on how that case should be handled (there's already a ticket).

apps/backend/fixtures/lambda/reconcile.json
fwkendall: Logged https://bcdevex.atlassian.net/browse/CCFPCM-618 to clean this up in future.

fwkendall (approved):

PR #205, opened 2023-08-16
[CHORE] Use current date in prod/test

chelsea-EYDS:

Objective: 
- use current date in both prod/test
- use inferred date if running the batch reconciliation, or on dev/local


chelsea-EYDS:
merging to the parent branch (not main) to avoid rebasing twice

PR #206, opened 2023-08-16
[CCFPCM-585] Sonarqube code smells

chelsea-EYDS:

Objective: 

- address the following "code smells" from sonarqube

Screenshot 2023-08-16 at 3 16 55 PM

Screenshot 2023-08-16 at 3 17 05 PM

Screenshot 2023-08-16 at 4 01 03 PM
Screenshot 2023-08-16 at 4 00 47 PM
Screenshot 2023-08-16 at 4 00 53 PM


apps/backend/src/common/entities/FixedWidthRecord.ts
```diff
@@ -96,7 +103,7 @@ export class FixedWidthRecord<T extends IFixedWidthRecord<T>>
     );
     // get `__proto__` and (recursively) all parent classes
     const rs = new Set([
-      ...(fields || []),
+      ...(fields ?? []),
```
fw-noel, line 106: This still works, but why the change?

apps/backend/src/common/entities/FixedWidthRecord.ts
```diff
@@ -132,9 +139,9 @@ export class FixedWidthRecord<T extends IFixedWidthRecord<T>>
     if (target.delimiter?.value) {
       target.delimiter.positions.forEach((pos: number) => {
         result = Buffer.concat([
-          result.slice(0, pos + 1),
+          result.subarray(0, pos + 1),
```
fw-noel, line 142: Why this change? I think it also works, but I'm curious

apps/backend/src/lambdas/pos-reconciliation.ts
```diff
@@ -0,0 +1,40 @@
+import { format, parse, subBusinessDays } from 'date-fns';
+import { DateRange, Ministries, NormalizedLocation } from '../constants';
+import { POSDepositEntity } from '../deposits/entities/pos-deposit.entity';
+import { PosReconciliationService } from '../reconciliation/pos-reconciliation.service';
+import { PaymentEntity } from '../transaction/entities';
+
+export const posReconciliationHandler = async (
```
fw-noel, line 7: Organization-wise, does this make sense in the lambdas section, or should it be maybe in a service?

fw-noel (approved):
I think this should all work, but I do have qs

apps/backend/src/common/entities/FixedWidthRecord.ts
fw-noel: Sorry nvm, just saw the smell

apps/backend/src/common/entities/FixedWidthRecord.ts
chelsea-EYDS: according to sonarqube slice is deprecated 🤷‍♀️

chelsea-EYDS:
converted to draft - found issues in the labour parsing - getting NaN instead of decimal value - will reopen when resolved

apps/backend/src/lambdas/pos-reconciliation.ts
chelsea-EYDS: resolved?

apps/backend/fixtures/lambda/reconcile.json
```diff
@@ -8,8 +8,8 @@
             "to": "2023-01-02"
           },
           "program": "SBC", 
-          "generateReport": true, 
-          "byPassFileValidity": false
+          "generateReport": false, 
+          "byPassFileValidity": true
```
fwkendall, line 12: I'm unclear how this change related to the SQ report?

apps/backend/src/common/entities/FixedWidthRecord.ts
```diff
@@ -61,23 +59,26 @@ export class FixedWidthRecord<T extends IFixedWidthRecord<T>>
...
+      switch (options?.format?.type) {
+        case DataType.Integer:
           (target as any)[field] = parseInt(value);
-        } else if (options.format.type === DataType.Float) {
-          (target as any)[field] = new Decimal(value)
-            .toDecimalPlaces(2)
-            .toNumber();
-        } else if (options.format.type === DataType.Date) {
+          break;
+        case DataType.Decimal:
```
fwkendall, line 67: So our old if/else was handling Float and not Decimal and the decimal.js docs are not helping me, will a Float pass this condition if passed in or does this effectively change behaviour?

apps/backend/src/lambdas/parser.ts
```diff
@@ -10,7 +10,7 @@ export const handler = async (event: S3Event, _context?: Context) => {
   appLogger.setContext('Parser Lambda');
   const parseService = app.get(ParseService);
   appLogger.log({ event, _context }, 'PARSE EVENT');
-  const automatedReconciliationEnabled = !process.env.DISABLE_LOCAL_RECONCILE;
+  const automatedReconciliationEnabled = process.env.RUNTIME_ENV === 'test';
```
fwkendall, line 13: What risks exist here - does this get set to "dev" or "prod" in those environments? Did we fully deprecate this var or could someone still try setting it and get unexpected behaviour?

apps/backend/src/lambdas/utils/eventTypes.ts
```diff
@@ -1,63 +1,66 @@
...
+      return 'isCloudfront';
 
-  if (event.configRuleId && event.configRuleName && event.configRuleArn)
-    return 'isAwsConfig';
+    case event.configRuleId && event.configRuleName && event.configRuleArn:
+      return 'isAwsConfig';
 
-  if (event.Records && event.Records[0].eventSource === 'aws:codecommit')
-    return 'isCodeCommit';
+    case event.Records && event.Records[0].eventSource === 'awscodecommit':
```
fwkendall, line 11: I see we removed the colon from all of these, was this broken before?

fwkendall: I'm pretty confused at the scope of event types we're triggering, there are quite a few services here that we don't have provisioned - do we have any history on whether lambda jobs actually trigger things like sqs?

apps/backend/src/common/entities/FixedWidthRecord.ts
chelsea-EYDS: Float was not actually being used anywhere so I removed it for clean up.

apps/backend/src/lambdas/utils/eventTypes.ts
chelsea-EYDS: This was not broken, but sonarqube complained about the complexity. We're not actually using this anywhere so it could be removed.

fwkendall: Logged [CCFPCM-617](https://bcdevex.atlassian.net/browse/CCFPCM-617)

apps/backend/src/lambdas/parser.ts
chelsea-EYDS: I have only set it to run on test currently as we don't have the prod env up yet. There is probably a better way to handle this one... we don't want it to run on local, and I've disabled it for dev so that Qudsia can run batch reconciliation on that env instead... maybe this warrants a discussion? Basically, if the sync job is running, or if we're in our production env, we want to enable automation, if the sync job is disabled, then we should run the data in batches...

chelsea-EYDS: I guess this should be in another ticket. The functionality is the same, I was just trying to provide clarity on where/when this should be disabled.

fwkendall: Well we're not in prod yet, so I'm thinking we don't need to effectively have a feature flag disabling this in prod right now? Otherwise we'll need to change this code again in order to perform the release unless I'm missing something.

chelsea-EYDS: This is meant to disable on dev. I see how I made it confusing. One moment I will fix it.

chelsea-EYDS: Screenshot 2023-08-22 at 10 26 12 AM

Screenshot 2023-08-22 at 10 26 40 AM

chelsea-EYDS: Setting it in the makefile to be more configurable...

fwkendall (approved):

PR #207, opened 2023-08-17
[CCFPCM-570] cron job for daily file check alerting

fw-noel:
[CCFPCM-570](https://bcdevex.atlassian.net/browse/CCFPCM-570)
[CCFPCM-612](https://bcdevex.atlassian.net/browse/CCFPCM-612)

Objective: 
- Automate alerts with a cron job using cloudwatch events at 11, 14, 17 (this may need to be adjusted depending on timezone)
- Only alert if files have been sent in across any LOB for the day
- We stop sending alerts if a daily upload was incomplete 7+ days ago. We can remove this logic; I just figure at that point, if files have continued to be uploaded, we can reconcile?
- Remove concept of "retries" as it's now in set times

### DIFF
--- a/apps/backend/src/database/migrations/1692293473030-migration.ts
+++ b/apps/backend/src/database/migrations/1692293473030-migration.ts
@@ -1,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migration1692293473030 implements MigrationInterface {
+  name = 'migration1692293473030';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "file_ingestion_rules" DROP COLUMN "retries"`
+    );
+    await queryRunner.query(
+      `ALTER TABLE "program_daily_upload" DROP COLUMN "retries"`
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "program_daily_upload" ADD "retries" integer NOT NULL`
+    );
+    await queryRunner.query(
+      `ALTER TABLE "file_ingestion_rules" ADD "retries" integer NOT NULL DEFAULT '0'`
+    );
+  }
+}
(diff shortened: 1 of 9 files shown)


apps/backend/src/database/migrations/1692293473030-migration.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migration1692293473030 implements MigrationInterface {
+  name = 'migration1692293473030';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
chelsea-EYDS, line 6: Is this migration going to run okay on test? Since we already data in this column?

apps/backend/src/lambdas/dailyfilecheck.ts
```diff
@@ -20,33 +19,54 @@ export const handler = async (event: unknown, context?: Context) => {
...
-  const date = format(new Date(), 'yyyy-MM-dd');
+  const date = new Date();
 
-  const dailyAlert: DailyAlertRO = await notificationService.dailyUploadAlert(
-    date
-  );
+  // Retrieve statuses from the past 7 days
+  const allRecentStatuses =
+    await notificationService.retrieveRecentDailyStatuses(
+      subDays(date, 7),
```
chelsea-EYDS, line 27: do we care about business days? (or does 7 normal days cover as you are only trying to look back 5?)

apps/backend/src/parse/parse.service.ts
```diff
@@ -473,20 +474,27 @@ export class ParseService {
...
       );
-
       if (!daily) {
-        daily = await this.notificationService.createNewDailyUploadRecord(
-          rules,
-          date
-        );
+        throw new Error('Error');
+      }
+      const missingFiles = await this.notificationService.findMissingDailyFiles(
```
chelsea-EYDS, line 490: sonarqube says we don't have to wait this one ?

chelsea-EYDS (dismissed):
Some minor comments, otherwise approved!

apps/backend/src/lambdas/dailyfilecheck.ts
fw-noel: I should remove this magic number. But essentially, yeah the buffer is pretty arbitrary, it's as if to say "enough days have passed and enough eyes should have seen this that we're going to continue reconciling if more recent files have been uploaded"

fw-noel: Updated

fwkendall (approved):
Looks great +1

PR #209, opened 2023-08-18
[FIX] api lambda added to webpack

chelsea-EYDS:

Objective: 

- added api lambda to webpack config
** note: this does not resolve the issue entirely as we are not building the swagger UI, but the endpoint is returning a 200 so this  is a partial fix


fwkendall (approved):

fwkendall:
Related to CCFPCM-616

PR #221, opened 2023-08-28
[CCFPCM-0433] enabled versioning (use makefile cmd to set version)

chelsea-EYDS:
[CCFPCM-0433](https://bcdevex.atlassian.net/browse/CCFPCM-0433)

Objective: 

- enable versioning
- update swagger module to make the requests to the proper endpoint
- remove global prefix and use enableVersioning 
- set version from package json
- include make commands to set version (can be used in a release flow/actions)
- use current version as default for all routes (this can be changed in the future if there is a need for version specific routing)


.yarnrc.yml
```diff
@@ -4,10 +4,12 @@ nodeLinker: node-modules
 
 plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
-    spec: '@yarnpkg/plugin-workspace-tools'
+    spec: "@yarnpkg/plugin-workspace-tools"
```
fw-noel, line 7: Why change these? Is this a linter thing?

fw-noel (commented):
This looks pretty good, but a rundown in the README would be a good idea

.yarnrc.yml
chelsea-EYDS: Yes - the pre-commit hook runs these so I don't see them in the diff prior to committing.

chelsea-EYDS:
(set to DRAFT - changes pending)

apps/backend/src/app.config.ts
```diff
@@ -96,8 +96,11 @@ export async function createNestApp(): Promise<{
   // Validation pipe
   app.useGlobalPipes(new TrimPipe(), new ValidationPipe(validationPipeConfig));
 
-  // Api prefix api/v1/
-  app.setGlobalPrefix(API_PREFIX);
+  app.enableVersioning({
+    type: VersioningType.URI,
+    prefix: 'api/v',
+    defaultVersion: process.env.APP_VERSION?.split('.')[0] ?? '1',
```
chelsea-EYDS, line 102: We only want to update the api path in the case of a major version update

apps/backend/src/common/documentation.ts
```diff
@@ -8,8 +8,8 @@ import { TransactionModule } from '../transaction/transaction.module';
 export const Documentation = (app: INestApplication) => {
   const options = new DocumentBuilder()
     .setTitle('PayCoCo API Docs')
+    .setVersion(process.env.APP_VERSION || '1.0.0')
```
chelsea-EYDS, line 11: Here we can show the actual version, but the path will be v2, v1, etc

package.json
```diff
@@ -9,6 +9,9 @@
     "node": "^18.14.0"
   },
   "scripts": {
+    "version:major": "yarn workspace @payment/backend version:major",
```
chelsea-EYDS, line 12: We can use these later as part of a release workflow? Running these cmds and then restarting the app will update the version displayed by swagger as well as the api paths (in the case of a major version update)

simensma-fresh (approved):
Looks great!

PR #223, opened 2023-08-28
[CCFPCM-407] Enabled s3 access logging

simensma-fresh:
[CCFPCM-407](https://bcdevex.atlassian.net/browse/CCFPCM-407)

Objective: 
- Enabled s3 access logging (https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerLogs.html)

Changes:
- Created new s3 bucket for storing S3 access logs (`pcc-s3-access-logs-${var.target_env}`)
- Enabled access logging on existing s3 buckets

### DIFF
--- a/terraform/s3.tf
+++ b/terraform/s3.tf
@@ -1,9 +1,20 @@
+locals {
+  pcc-reporting-bucket-name = "pcc-recon-reports-${var.target_env}"
+  pcc-master-data-bucket-name = "pcc-master-data-${var.target_env}"
+  pcc-deployments-bucket-name = "pcc-deployments-${var.target_env}"
+}
+
 
 resource "aws_s3_bucket" "pcc-reporting" {
-  bucket = "pcc-recon-reports-${var.target_env}"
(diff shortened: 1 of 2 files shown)


fwkendall (approved):
Looks good! 👍

terraform/s3.tf
```diff
@@ -1,10 +1,21 @@
+locals {
+  pcc-reporting-bucket-name = "pcc-recon-reports-${var.target_env}"
```
chelsea-EYDS, line 2: we also have the "pcc-integration-data-files-${var.target_env}" bucket (we drop the files from our sftp server via aws-xfer family here). I am not super clear on how locals work in terraform but I just wanted to make sure that we didn't need to include this bucket as well?

simensma-fresh: @chelsea-EYDS  I set up the logging of that bucket separately (in sftp.tf) so, should not be needed here

chelsea-EYDS (approved):

PR #224, opened 2023-08-28
[CCFPCM-407] Disabled rds backups in dev, increased retention period in test to match prod

simensma-fresh:
[CCFPCM-407](https://bcdevex.atlassian.net/browse/CCFPCM-407)

Objective: 
- Disabled DB backups in dev environment
- Increased DB backup retention period in test environment to match production (14 days). This will also resolve the DB backup not enabled alert in security hub which required a minimum 7 day retention period.

### DIFF
--- a/terraform/db.tf
+++ b/terraform/db.tf
@@ -31,7 +31,7 @@ resource "aws_rds_cluster" "pgsql" {
 
   # 2AM-4AM PST
   preferred_backup_window = "09:00-11:00"
-  backup_retention_period = var.target_env == "prod" ? 14 : 3
+  backup_retention_period = var.target_env == "dev" ? 0: 14
 
   lifecycle {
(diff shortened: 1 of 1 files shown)


terraform/db.tf
```diff
@@ -31,7 +31,7 @@ resource "aws_rds_cluster" "pgsql" {
 
   # 2AM-4AM PST
   preferred_backup_window = "09:00-11:00"
-  backup_retention_period = var.target_env == "prod" ? 14 : 3
+  backup_retention_period = var.target_env == "dev" ? 0: 14
```
fwkendall, line 34: @chelsea-EYDS just for your attention, in case you look to spin up the tools env next week - this will also need tweaking 😄

fwkendall (approved):
Looks good! I think test and prod staying as close as possible in config makes a lot of sense. 👍

chelsea-EYDS (approved):

PR #239, opened 2023-09-11
[FIX] reconciliation filecheck date

chelsea-EYDS:

Objective: 

- file check date should be minus one business day

### DIFF
--- a/apps/backend/fixtures/lambda/reconcile.json
+++ b/apps/backend/fixtures/lambda/reconcile.json
@@ -3,7 +3,7 @@
     {
       "Sns": {
         "Message": {
-          "reconciliationMaxDate": "2023-08-14",
+          "reconciliationMaxDate": "2023-09-11",
           "program": "SBC",
           "reportEnabled": true,
           "byPassFileValidity": false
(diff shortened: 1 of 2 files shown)


fw-noel (approved):
I see now what you were saying, and yeah you were right.

I'm wary in general of using subBusinessDays because of holidays, but it should be relatively harmless here because ultimately if there's no data it won't reconcile. The one thing I'm thinking about is the way the logic works in daily file check, where we use `created_at` to see if a status was created today, which actually doesn't work because of the SBC file getting uploaded at night. I wanted to consider consistency, but I don't think there's a possible way to do the same logic on both ends right now.

chelsea-EYDS:
> I see now what you were saying, and yeah you were right.
> 
> I'm wary in general of using subBusinessDays because of holidays, but it should be relatively harmless here because ultimately if there's no data it won't reconcile. The one thing I'm thinking about is the way the logic works in daily file check, where we use `created_at` to see if a status was created today, which actually doesn't work because of the SBC file getting uploaded at night. I wanted to consider consistency, but I don't think there's a possible way to do the same logic on both ends right now.

For the daily file check, can we use the metadata date instead instead of the created at? Since we know that the data and metadata date from an SBC Garms file corresponds to a TDI file in which the metadata date (not the created at date) then we should be able to run based on this? So, we will have all three files as of 11am ish when we get the last TDI file?

PR #245, opened 2023-09-18
[CCFPCM-666] Updated webpack for the wipe db lambda

chelsea-EYDS:
[CCFPCM-0666](https://bcdevex.atlassian.net/browse/CCFPCM-0666)

Objective: 

- Updated webpack config to include the wipe dev db lambda

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,6 +225,7 @@ aws-run-clear-dev-data:
 	@touch clear-db-results || true
 	@aws lambda invoke --function-name clearDevData --payload '{}' clear-db-results  --region ca-central-1
 	@cat clear-db-results | grep "success"
+	@rm clear-db-results
 
 aws-run-reconciler:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -222,7 +222,7 @@ aws-run-migrator:
 	@cat migration-results | grep "success"
 
 aws-run-clear-dev-data: 
-	@rm clear-db-results || true
+	@touch clear-db-results || true
```
fw-noel, line 225: These are very different, what's the reasoning?

chelsea-EYDS: We need a file to output the results to. I kept seeing a message that the rm <filename> couldn't be found. (We see it in the migrations as well). I guess I could delete the file afterwards as well?

chelsea-EYDS: Okay updated to also remove the file afterwards. This doesn't change the functionality but it removes an un-needed warning message about the file not existing.

fw-noel (approved):

PR #249, opened 2023-09-22
[CCFPCM-0667] sync job 

chelsea-EYDS:
[CCFPCM-0667](https://bcdevex.atlassian.net/browse/CCFPCM-0667)

Objective: 

- update bucket policies for pcc-data-files to allow for sync to run

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,8 @@ aws-build-and-deploy-aws-data-clear-lambda: build-backend aws-upload-artifacts a
 # ======================================================================
 
 aws-sync-files-from-prod-to-dev:
-	@aws s3 sync s3://pcc-integration-data-files-prod s3://pcc-integration-data-files-dev --acl bucket-owner-full-control
+	@aws s3 sync s3://pcc-integration-data-files-prod/sbc s3://pcc-integration-data-files-dev/sbc --acl bucket-owner-full-control
+	@aws s3 sync s3://pcc-integration-data-files-prod/bcm s3://pcc-integration-data-files-dev/bcm --acl bucket-owner-full-control
 
 aws-sync-files-from-prod-to-test:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -213,11 +213,15 @@ aws-build-and-deploy-aws-data-clear-lambda: build-backend aws-upload-artifacts a
 # AWS Interactions
 # ======================================================================
 
+	
+
 aws-sync-files-from-prod-to-dev:
-	@aws s3 sync s3://pcc-integration-data-files-prod s3://pcc-integration-data-files-dev --acl bucket-owner-full-control
+	@aws s3 sync s3://pcc-integration-data-files-prod/sbc s3://pcc-integration-data-files-dev/sbc --acl bucket-owner-full-control
```
chelsea-EYDS, line 219: This change is not related to the bug, but it prevents the archived directory from syncing (this causes an error to be printed due to an issue with a file in our archive)

simensma-fresh (approved):

Makefile
fwkendall: Seems reasonable not to sync the archive - we might want to mention this to Qudsia too just in case, but if we need to sync that we can make a new ticket for it 😄

terraform/s3.tf
```diff
@@ -9,7 +9,6 @@ locals {
     aws_s3_bucket.pcc-master-data,
     aws_s3_bucket.pcc-deployments,
     aws_s3_bucket.pcc-s3-access-logs,
-    aws_s3_bucket.sftp_storage,
```
fwkendall, line 12: I take it this was just fully redundant with sftp.tf ?

fwkendall (approved):

terraform/s3.tf
chelsea-EYDS: From what I could tell the policies set to each bucket in this list were overriding the inline policy set from the sftp.tf file for this bucket which allows the sync job to run.

chelsea-EYDS:
Merging into the base branch (not main)

PR #250, opened 2023-09-27
[CCFPCM-0679] Bugfix for manual parsing 

chelsea-EYDS:
[CCFPCM-0679](https://bcdevex.atlassian.net/browse/CCFPCM-0679)

Objective: 
- The data bucket has too many items to list in a single call when manually parsing. This does not affected automated parsing as we are simply parsing the files dropped in through the automation flow, vs comparing the filelist from s3 with the db file upload table. 

- remove archive from dev/local and do not sync the archive folder


fw-noel (approved):
I think this is fine, I might want a command to pull in archives as well. To get testing data.

PR #251, opened 2023-09-28
[CCFPCM-672] add errors to limit file extensions

fw-noel:
[CCFPCM-0000](https://bcdevex.atlassian.net/browse/CCFPCM-627)

Objective: 
Limit file extensions coming in on the API endpoint

### DIFF
--- a/apps/backend/src/parse/parse.controller.ts
+++ b/apps/backend/src/parse/parse.controller.ts
@@ -7,12 +7,12 @@ import {
   ClassSerializerInterceptor,
   Inject,
   HttpException,
+  BadRequestException,
 } from '@nestjs/common';
 import { FileInterceptor } from '@nestjs/platform-express';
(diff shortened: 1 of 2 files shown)


apps/backend/src/parse/parse.controller.ts
```diff
@@ -69,6 +69,23 @@ export class ParseController {
     @Body() body: { program: Ministries; fileType: FileTypes },
     @UploadedFile() file: Express.Multer.File
   ) {
+    const fileSplit = file.filename.split('.');
+    const fileExtension = fileSplit[fileSplit.length - 1];
+    if (
+      fileSplit.length < 2 ||
```
chelsea-EYDS, line 75: This looks good... I know that multer has some options that can be configured for validation as well - I think mimetype is one of them... although I am not sure re: DAT/JSON but I think you can specify 🤷‍♀️

chelsea-EYDS (approved):

PR #254, opened 2023-10-03
[CCFPCM-654] File Upload

chelsea-EYDS:
[CCFPCM-0654](https://bcdevex.atlassian.net/browse/CCFPCM-0654)

Objective: 

File upload endpoint accepts supported file types (metadata and extension)

The uploaded file is stored in the appropriate S3 bucket for its LOB and file type

The parser lambda is triggered immediately for the file when it appears in S3

The API endpoint returns a 20x success code if the file could be added to S3, otherwise an appropriate error code depending what went wrong

When a file is uploaded that is already present in S3, this endpoint will still return a success message as it replaces the file in S3 (parsing validation is done separately)

Swagger docs are updated

### DIFF
--- a/apps/backend/src/constants.ts
+++ b/apps/backend/src/constants.ts
@@ -8,7 +8,7 @@ import { PaymentEntity, PaymentMethodEntity } from './transaction/entities';
 
 enum FileExtensions {
   DAT = 'DAT',
-  JSON = 'JSON"',
+  JSON = 'JSON',
 }
 
 export const ALL = 'all';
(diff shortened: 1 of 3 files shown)


apps/backend/src/constants.ts
```diff
@@ -8,7 +8,7 @@ import { PaymentEntity, PaymentMethodEntity } from './transaction/entities';
 
 enum FileExtensions {
   DAT = 'DAT',
-  JSON = 'JSON"',
+  JSON = 'JSON',
```
fw-noel, line 11: Oh boy, thank you

fw-noel (approved):

PR #255, opened 2023-10-04
[FIX] POS Round Three

chelsea-EYDS:

Objective: 

- Found an issue, round three heuristic rules were not being properly applied
- negative amounts (between payments and deposits) also resolved
sample report: 
(you can clearly see where a more generalized heuristics round to match many to many across multiple days would increase the match % - for round three we are matching one to one, hence the exceptions (see Campbell River for example) 

[reconciliation_report_2023-09-12-2023-09-30.xlsx](https://github.com/bcgov/PaymentCommonComponent/files/12822963/reconciliation_report_2023-09-12-2023-09-30.xlsx)


apps/backend/src/lambdas/reconcile.ts
```diff
@@ -32,7 +37,7 @@ export const handler = async (event: SNSEvent, _context?: Context) => {
   const notificationService = app.get(NotificationService);
   const snsService = app.get(SnsManagerService);
   const reportingService = app.get(ReportingService);
-  const numDaysToReconcile = 31;
+  const numDaysToReconcile = 14;
```
fw-noel, line 40: I understand this difference probably isn't meaningful in the grand scheme, but can I ask **why** this is a needed change?

apps/backend/src/lambdas/reconcile.ts
```diff
@@ -219,11 +142,139 @@ export const handler = async (event: SNSEvent, _context?: Context) => {
...
+      appLogger
+    );
     reportEnabled && (await generateReport());
     isLocal && (await showConsoleReport());
   } catch (err) {
     appLogger.error(err);
   }
 };
+
+const findPosExceptions = async (
```
fw-noel, line 177: Comments above each of these with expected outcomes please!

fw-noel (commented):
Do we have sufficient test coverage for `reconcilePos` and `reconcileCash`?

apps/backend/src/lambdas/reconcile.ts
chelsea-EYDS: (I do think we should decrease the no. of days we're reconciling for but will need the team/Aiden to agree :) )

chelsea-EYDS:
> Do we have sufficient test coverage for `reconcilePos` and `reconcileCash`? If not, would love that in a different PR

These are the same functions just moved outside of the lambda helper for readability so I have not changed or updated any of the tests, but I do not think we have any tests specifically for the lambda. I will make a tech debt ticket to test these :)

fw-noel (approved):

PR #260, opened 2023-10-12
[FIX] report lamba memory issue (#259)

chelsea-EYDS:

Objective: 

-


apps/backend/src/lambdas/reconcile.ts
```diff
@@ -51,7 +51,7 @@ export const handler = async (event: SNSEvent, _context?: Context) => {
   const currentDate = parse(reconciliationMaxDate, 'yyyy-MM-dd', new Date());
 
   const reconciliationMinDate = format(
-    subBusinessDays(currentDate, numDaysToReconcile),
+    subDays(currentDate, numDaysToReconcile),
```
fw-noel, line 54: Why is this change a part of it?

fw-noel (approved):

PR #272, opened 2023-10-27
[CCFPCM-0426] PART 1

chelsea-EYDS:
[CCFPCM-0426](https://bcdevex.atlassian.net/browse/CCFPCM-0426)

Objective:

- normalize the existing locations table by splitting into locations, merchants, banks
- set up seed script to ease the process for future ministry's to onboard (will require some - update to the logic as well as the new minstry's location data, but the script will check for the locations per ministry and we can insert new ministry data if none is found).
- 
[DB Diagram](https://dbdocs.io/chelsea-EYDS/bcpcc?view=relationships)
(password: paycoco)

NOTE: added do not merge label - we have a location id in our prod database that does not exit in our master location data and so the new migrations will fail until this location is added into the db


apps/backend/src/location/entities/location.entity.ts
```diff
@@ -0,0 +1,23 @@
+import { Entity, Unique, OneToMany, Relation, Index } from 'typeorm';
+import { BaseLocationEntity } from './location_base';
+import { BankLocationEntity, MerchantEntity } from '.';
+
+@Entity('location')
+@Unique(['id', 'location_id', 'source_id'])
+@Index('location_source_idx', ['id', 'location_id', 'source_id'], {
```
fw-noel, line 7: Brought up in chat: i don't think id should be part of this index

apps/backend/src/location/entities/master-location-data.entity.ts
```diff
@@ -1,50 +1,17 @@
-import { Entity, Column, PrimaryGeneratedColumn } from 'typeorm';
+import { Column, Entity } from 'typeorm';
+import { BaseLocationEntity } from './location_base';
 
+//TODO rename
```
fw-noel, line 4: todo

fw-noel (approved):

PR #274, opened 2023-10-30
[CCFPCM-0426] PART 2 db normalization

chelsea-EYDS:
[CCFPCM-0426](https://bcdevex.atlassian.net/browse/CCFPCM-0426)

Objective:


apps/backend/src/location/entities/location.entity.ts
```diff
@@ -10,11 +10,13 @@ import { BankLocationEntity, MerchantEntity } from '.';
 export class MinistryLocationEntity extends BaseLocationEntity {
   @OneToMany(() => BankLocationEntity, (bank) => bank.location, {
     cascade: true,
+    eager: true,
```
fw-noel, line 13: Please justify these eagers

apps/backend/src/location/location.service.ts
```diff
@@ -99,56 +98,22 @@ export class LocationService {
...
     await this.locationRepo.save(this.locationRepo.create(locationsData));
   }
 
   public async getLocationsByID(
     program: Ministries,
     location_ids: number[]
-  ): Promise<NormalizedLocation[]> {
-    const locations = await this.locationRepo.find({
+  ): Promise<MinistryLocationEntity[]> {
+    return await this.ministryLocationRepo.find({
```
fw-noel, line 116: If you're returning straight, I'm pretty sure you don't need the await, as `.find` is already an async function whose results will be properly returned

apps/backend/test/mocks/classes/cash_deposit_mock.ts
```diff
@@ -19,7 +19,7 @@ export class CashDepositMock extends CashDepositEntity {
     this.deposit_date = dateRange.maxDate;
     this.metadata = metadata;
     this.deposit_amt_cdn = amount;
-    this.pt_location_id = location.pt_location_id;
+    this.pt_location_id = location.banks[0].bank_id;
```
fw-noel, line 22: Is this always true that pt_location_id corresponds to this? A location won't always have a set of banks right?

I know this is a mock, but I'm curious

apps/backend/src/location/entities/location.entity.ts
chelsea-EYDS: We almost always need these together - at least to integrate with the code easily. Could be removed later if we refactor.

apps/backend/test/mocks/classes/cash_deposit_mock.ts
chelsea-EYDS: Yes, a location always has a set of banks.

apps/backend/src/location/entities/location.entity.ts
chelsea-EYDS: I removed the eager from pos deposit and cash deposit, in the case of locations though we need the banks/merchants in most cases

chelsea-EYDS: ...not sure if my response is sufficient - please let me know if there is any action needed to resolve 😁

apps/backend/test/mocks/classes/cash_deposit_mock.ts
chelsea-EYDS: Hmm, you are correct though, not all banks will necessarily have an entry in the cash deposit files. I am wondering if we should include the method on this table.
(all banks have a location_id, all location_ids have banks, all cash_deposit rows have a pt_location_id which corresponds to a bank_id which corresponds to a location, but since we're storing the bank_id for non cash as well then it's possible that banks[0] would not correspond to a cash type bank_id.)

fw-noel (dismissed):

fw-noel (approved):

PR #282, opened 2023-11-02
[CCFPCM-700] fix: json parsing error message

chelsea-EYDS:
[CCFPCM-700](https://bcdevex.atlassian.net/browse/CCFPCM-700)

Objective: 

- ensure the error message sends the alert and is clear on the error for malformed JSON

### DIFF
--- a/apps/backend/src/parse/parse.service.ts
+++ b/apps/backend/src/parse/parse.service.ts
@@ -80,7 +80,7 @@ export class ParseService {
     errantColumnName: string,
     errantIdColumnName: string
   ): string {
-    let errorMessage = `Error parsing ${fileName}. Please ensure all rows are valid.`;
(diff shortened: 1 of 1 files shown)


apps/backend/src/parse/parse.service.ts
```diff
@@ -112,29 +111,35 @@ export class ParseService {
    */
   async parseGarmsFile(
     contents: string,
-    fileName: string,
-    locations: MinistryLocationEntity[]
+    fileName: string
   ): Promise<{ txnFile: TransactionEntity[]; txnFileDate: string }> {
     const paymentMethods = await this.paymentMethodService.getPaymentMethods();
+    const locations = await this.locationService.findAllMinistryLocations();
```
chelsea-EYDS, line 117: Moved the locations lookup within each function (this way we can return the parsed data without passing in the locations - easier for logging if files are valid)

fw-noel (approved):

PR #289, opened 2023-11-17
[CCFPCM-703] fix: increase lambda memory

chelsea-EYDS:
[CCFPCM-0703](https://bcdevex.atlassian.net/browse/CCFPCM-0703)

Objective: 

- didn't push this change on the last one!


fw-noel (approved):
I'm approving, but any issues halving this? Seems like a drastic increase.

